PR #5, opened 2017-10-20
Fix floating point parsing on non-English Windows

paavohuhtala:
`float.Parse` uses the current locale setting of the OS to decide which floating point format to use. This is a problem in countries which use the [decimal comma](https://en.wikipedia.org/wiki/Decimal_mark#Countries_using_Arabic_numerals_with_decimal_comma) (which include the vast majority of European and South American countries).

This PR fixes that issue by making `InvariantCulture` (basically `en-us`) the default culture for the main thread and newly created threads.

### DIFF
--- a/src/OpenSage.Game/Data/Utilities/ParseUtility.cs
+++ b/src/OpenSage.Game/Data/Utilities/ParseUtility.cs
@@ -1,0 +1,17 @@
+﻿using System.Globalization;
+
+namespace OpenSage.Data.Utilities
+{
+    internal static class ParseUtility
+    {
+        public static float ParseFloat(string s)
+        {
+            return float.Parse(s, CultureInfo.InvariantCulture);
+        }
+
+        public static bool TryParseFloat(string s, out float result)
+        {
+            return float.TryParse(s, out result);
+        }
+    }
+}
--- a/src/OpenSage.Game/Data/Utilities/ParseUtility.cs
+++ b/src/OpenSage.Game/Data/Utilities/ParseUtility.cs
@@ -11,7 +11,7 @@ namespace OpenSage.Data.Utilities
 
         public static bool TryParseFloat(string s, out float result)
         {
-            return float.TryParse(s, out result);
+            return float.TryParse(s, NumberStyles.Float, CultureInfo.InvariantCulture, out result);
         }
     }
 }
--- a/src/OpenSage.Game/Data/W3d/W3dVertexMapperArgs.cs
+++ b/src/OpenSage.Game/Data/W3d/W3dVertexMapperArgs.cs
@@ -86,11 +86,11 @@ namespace OpenSage.Data.W3d
                 switch (mapperArgName)
                 {
                     case "UPerSec":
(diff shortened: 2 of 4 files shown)


paavohuhtala:
I'm now able to run the tests (#6), and I've come to realize this change needs to be done for them too.

xUnit.NET doesn't provide a built-in way to run tests using a specific culture, or a way to have "global" initialization code run before every test.

Maybe a better idea would be to change the parsing code directly, by wrapping `float.Parse` into a new method which would use `CultureInfo.InvariantCulture`?

tgjones:
Thank you for this!

I agree that it would be better to change the parsing code directly. As you suggested, let's add a new wrapper method that calls `float.Parse` with `CultureInfo.InvariantCulture`.

I don't really want to change the main thread's culture, because later on it would be nice to support properly localized Generals resources using the main thread culture as the locale.

src/OpenSage.Game/Data/Utilities/ParseUtility.cs
```diff
@@ -0,0 +1,17 @@
...
+    internal static class ParseUtility
+    {
+        public static float ParseFloat(string s)
+        {
+            return float.Parse(s, CultureInfo.InvariantCulture);
+        }
+
+        public static bool TryParseFloat(string s, out float result)
+        {
+            return float.TryParse(s, out result);
```
tgjones: Shouldn't this use the `TryParse` overload that you can pass a culture to?

paavohuhtala: Oops.

src/OpenSage.Game/Data/W3d/W3dVertexMapperArgs.cs
```diff
@@ -85,23 +86,23 @@ public static W3dVertexMapperArgs Parse(string value)
...
                         break;
 
                     case "VPerSec":
-                        result.VPerSec = float.Parse(mapperArgValue);
+                        result.VPerSec = ParseFloat(mapperArgValue);
                         break;
 
                     case "UScale":
-                        float.TryParse(mapperArgValue, out result.UScale);
+                        TryParseFloat(mapperArgValue, out result.UScale);
```
paavohuhtala: I was wondering - `float.TryParse` is used twice this way in this file, but the return value is never checked.

Is this because these values might be corrupted / invalid but it's not a critical issue, and in that case you want to set the out variable to 0?

tgjones: Yes, some of them contain invalid data (specifically, some have no value but are missing the comment character - `;` - before the comment).

Actually all of the cases could / should use `TryParseFloat`, but so far I only changed it for the ones I needed to for the actual data.

paavohuhtala: I did some regex-fu and replaced all instances with `TryParseFloat`.

tgjones:
Thank you!

PR #6, opened 2017-10-20
WIP: Resolve game installation paths from Windows registry

### LINKED ISSUE #4: Unit tests use hard coded paths
_You are most likely already aware of this, but I think it's a good idea to write it down regardless._

The tests use hard coded paths to game files, which makes it hard to run them portably. Most tests seem to use `InstalledFilesTestData`, but some other tests (such as the `BigArchiveTests`) use their own paths.

There are a couple of ways to solve this.

The easiest to implement solution would be to have a configuration file which a developer would enter the path[s] to the game installation folder[s].

From developer experience perspective an even better solution would be to automatically detect where the game is installed. (Much) further down the line, the same logic could be used for the game itself. 

There are at least 3 different editions (original, The First Decade and Origin) of the game, which might require different approaches. I have the Origin copy, and having done some preliminary Windows registry searching and `%appdata%` grepping I couldn't find any non-machine specific references to the game installation path. There must be a way to do this, but I haven't found it yet. :)

paavohuhtala:
This is a work in progress implementation of a registry-based installation folder resolver, as mentioned in #4. The core feature is functional, and I'm now able to run the data viewer on my computer and access the data from Origin copies of Generals & Zero Hour. There is still some work to be done - see the checklist below and the self-review comments.

- [x] Use an installation locator for tests (resolves #4)
- [ ] Add registry keys for BFME & BFME 2

### DIFF
--- a/src/OpenSage.DataViewer/ViewModels/ShellViewModel.cs
+++ b/src/OpenSage.DataViewer/ViewModels/ShellViewModel.cs
@@ -1,10 +1,9 @@
-﻿using System;
-using System.Collections.Generic;
-using System.IO;
+﻿using System.Collections.Generic;
 using System.Linq;
 using System.Windows.Input;
 using Caliburn.Micro;
 using OpenSage.DataViewer.Framework;
+using OpenSage.Data;
 
 namespace OpenSage.DataViewer.ViewModels
 {
@@ -38,18 +37,11 @@ namespace OpenSage.DataViewer.ViewModels
 
         private static IEnumerable<InstallationViewModel> FindInstallations()
         {
-            var programFilesPath = Environment.GetFolderPath(Environment.SpecialFolder.ProgramFilesX86);
+            var locator = IoC.Get<IInstallationLocator>();
 
-            var possibleInstallations = new[]
-            {
-                new InstallationViewModel("C&C Generals (The First Decade)", Path.Combine(programFilesPath, @"EA Games\Command & Conquer The First Decade\Command & Conquer(tm) Generals")),
-                new InstallationViewModel("C&C Generals (Origin)", Path.Combine(programFilesPath, @"Origin Games\Command and Conquer Generals Zero Hour\Command and Conquer Generals")),
-                new InstallationViewModel("C&C Generals Zero Hour (The First Decade)", Path.Combine(programFilesPath, @"EA Games\Command & Conquer The First Decade\Command & Conquer(tm) Generals Zero Hour")),
-                new InstallationViewModel("C&C Generals Zero Hour (Origin)", Path.Combine(programFilesPath, @"Origin Games\Command and Conquer Generals Zero Hour\Command and Conquer Generals Zero Hour")),
-                new InstallationViewModel("Battle for Middle-earth (DVD)", Path.Combine(programFilesPath, @"EA Games\The Battle for Middle-earth (tm)"))
-            };
-
-            return possibleInstallations.Where(x => Directory.Exists(x.Path));
+            return Games.GetAll()
+                .SelectMany(locator.FindInstallations)
+                .Select(installation => new InstallationViewModel(installation.DisplayName, installation.Path));
         }
     }
 }
--- a/src/OpenSage.DataViewer/ViewModels/ShellViewModel.cs
+++ b/src/OpenSage.DataViewer/ViewModels/ShellViewModel.cs
@@ -39,7 +39,7 @@ namespace OpenSage.DataViewer.ViewModels
         {
             var locator = IoC.Get<IInstallationLocator>();
 
-            return Games.GetAll()
+            return SageGames.GetAll()
                 .SelectMany(locator.FindInstallations)
                 .Select(installation => new InstallationViewModel(installation.DisplayName, installation.Path));
         }
--- a/src/OpenSage.Game.Tests/InstalledFilesTestData.cs
+++ b/src/OpenSage.Game.Tests/InstalledFilesTestData.cs
@@ -14,7 +14,7 @@ namespace OpenSage.Data.Tests
             Locator = new RegistryInstallationLocator();
         }
 
-        public static string GetInstallationDirectory(SageGame game) => Locator.FindInstallations(game).FirstOrDefault().Path;
+        public static string GetInstallationDirectory(SageGame game) => Locator.FindInstallations(game).First().Path;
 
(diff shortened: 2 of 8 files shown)


src/OpenSage.Game/Data/InstallationLocator.cs
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    public interface IInstallationLocator
+    {
+        IEnumerable<GameInstallation> FindInstallations(GameId game);
+    }
+
+    public class RegistryInstallationLocator : IInstallationLocator
+    {
+        private static readonly (string, string)[] GeneralsKeys = new[] { (@"SOFTWARE\Electronic Arts\EA Games\Generals", "InstallPath") };
```
paavohuhtala: These are arrays, because games could have different editions / releases with different registry keys. If that turns out not to be the case, we can just store single keys.

src/OpenSage.Game/Data/InstallationLocator.cs
```diff
@@ -0,0 +1,98 @@
...
+        private static readonly (string, string)[] ZeroHourKeys = new[] { (@"SOFTWARE\Electronic Arts\EA Games\Command and Conquer Generals Zero Hour", "InstallPath") };
+
+        private IEnumerable<(string keyName, string valueName)> GetRegistryKeysForGame(GameId game)
+        {
+            switch (game)
```
paavohuhtala: This could be a hashmap, but with lists this small a switch is faster and supports exhaustiveness checking.

src/OpenSage.DataViewer/ViewModels/ShellViewModel.cs
```diff
@@ -38,18 +37,11 @@ public ShellViewModel()
...
-                new InstallationViewModel("C&C Generals (Origin)", Path.Combine(programFilesPath, @"Origin Games\Command and Conquer Generals Zero Hour\Command and Conquer Generals")),
-                new InstallationViewModel("C&C Generals Zero Hour (The First Decade)", Path.Combine(programFilesPath, @"EA Games\Command & Conquer The First Decade\Command & Conquer(tm) Generals Zero Hour")),
-                new InstallationViewModel("C&C Generals Zero Hour (Origin)", Path.Combine(programFilesPath, @"Origin Games\Command and Conquer Generals Zero Hour\Command and Conquer Generals Zero Hour")),
-                new InstallationViewModel("Battle for Middle-earth (DVD)", Path.Combine(programFilesPath, @"EA Games\The Battle for Middle-earth (tm)"))
-            };
-
-            return possibleInstallations.Where(x => Directory.Exists(x.Path));
+            return Games.GetAll()
+                .SelectMany(locator.FindInstallations)
+                .Select(installation => new InstallationViewModel(installation.DisplayName, installation.Path));
```
paavohuhtala: There could be multiple different installations of the same game, and there is currently no way to distinguish between them in the UI.

src/OpenSage.Game/Data/InstallationLocator.cs
```diff
@@ -0,0 +1,98 @@
+﻿using Microsoft.Win32;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace OpenSage.Data
+{
+    public enum GameId
```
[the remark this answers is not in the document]
paavohuhtala: Having read more of the code, re-using `OpenSage.Data.Ini.SageGame` might make sense here? It could also be moved higher in the namespace hierarchy if it's commonly used.

src/OpenSage.Game.Tests/InstalledFilesTestData.cs
```diff
@@ -7,15 +7,18 @@ namespace OpenSage.Data.Tests
 {
     internal static class InstalledFilesTestData
     {
+        private static readonly IInstallationLocator locator;
```
tgjones: It's a minor thing, but please could you follow the naming convention I've used in the rest of the code, and name this `Locator`.

src/OpenSage.Game/Data/InstallationLocator.cs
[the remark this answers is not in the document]
tgjones: Yes - let's move `SageGame` to the root of the OpenSage.Game project, in the `OpenSage` namespace.

src/OpenSage.Game/Data/InstallationLocator.cs
```diff
@@ -0,0 +1,98 @@
...
+        public static IEnumerable<GameId> GetAll()
+        {
+            yield return GameId.Generals;
+            yield return GameId.ZeroHour;
+        }
+    }
+
+    public interface IInstallationLocator
+    {
+        IEnumerable<GameInstallation> FindInstallations(GameId game);
```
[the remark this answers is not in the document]
tgjones: That sounds fine - I don't plan to support UWP any time soon, so we're stuck with blocking file system APIs anyway.

tgjones:
This is great, thank you! Please could you have a look at my comments, and then this will be ready to merge.

At some point soon we're going to have to add the extra logic for Zero Hour, where it looks in the Zero Hour folder and .big files first, and then falls back to the Generals folder and .big files. But that can be done separately.

paavohuhtala:
@tgjones I've done the requested changes.

src/OpenSage.Game.Tests/InstalledFilesTestData.cs
```diff
@@ -7,18 +7,18 @@ namespace OpenSage.Data.Tests
...
+        private static readonly IInstallationLocator Locator;
 
         static InstalledFilesTestData()
         {
-            locator = new RegistryInstallationLocator();
+            Locator = new RegistryInstallationLocator();
         }
 
-        public static string GetInstallationDirectory(GameId game) => locator.FindInstallations(game).FirstOrDefault().Path;
+        public static string GetInstallationDirectory(SageGame game) => Locator.FindInstallations(game).FirstOrDefault().Path;
```
tgjones: I think this should be `FirstOrDefault()?.Path`.

paavohuhtala: I can change it, although it won't make a _huge_ difference - the test will crash anyway if the game files can't be found, but the stack trace will point to the usage site of the return value instead of here.

tgjones: In that case, could you change it to `First().Path`, then it will still crash, but with a slightly more meaningful exception :)

tgjones:
Merged, thank you!

PR #12, opened 2017-12-23
Initial apt parsing

feliwir:
Added most of the basic character parsing and added two tests, to verify that apt loading is working as expected


src/OpenSage.Game/Data/Apt/AptFile.cs
```diff
@@ -0,0 +1,88 @@
...
+        internal FileSystem FileSystem;
+
+        public AptFile(ConstantData constants, FileSystem filesystem, string name)
+        {
+            Constants = constants;
+            FileSystem = filesystem;
+            MovieName = name;
+        }
+
+        public void Parse(BinaryReader reader)
```
tgjones: Could this be a private method? I don't think it's called from anywhere other than `AptFile.FromFileSystemEntry`.

src/OpenSage.Game/Data/Apt/AptFile.cs
```diff
@@ -0,0 +1,88 @@
...
+{
+    public class AptFile
+    {
+        public ConstantData Constants { get; private set; }
+        public Movie Movie { get; private set; }
+        internal bool IsEmpty = true;
+        internal string MovieName;
+        internal FileSystem FileSystem;
+
+        public AptFile(ConstantData constants, FileSystem filesystem, string name)
```
tgjones: Can this be a `private` constructor?

src/OpenSage.Game/Data/Apt/AptFile.cs
```diff
@@ -0,0 +1,88 @@
...
+                AptFile importApt;
+
+                if (importDict.ContainsKey(import.Movie))
+                {
+                    importApt = importDict[import.Movie];
+                }
+                else
+                {
+                    var importEntry = FileSystem.GetFile(Path.ChangeExtension(import.Movie, ".apt"));
+                    importApt = AptFile.FromFileSystemEntry(importEntry);
```
tgjones: Presumably this could result in parsing an `.apt` file multiple times? That's why I was thinking this resolving of imports / exports would happen outside `AptFile`, in a separate layer that can make sure `.apt` files only get parsed once. What do you think?

src/OpenSage.Game/Data/Apt/Characters/Button.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    public class Button : Character
+    {
+        public Vector4 Bounds { get; private set; }
+        public IndexedTriangle[] Triangles { get; private set; }
+        public Vector2[] Vertices { get; private set; }
+
+        public static Button Parse(BinaryReader reader)
+        {
+            var b = new Button();
```
tgjones: I'd prefer if you used full words for local variable names, like `button` here. I know there are several different schools of thought on this, but that's what I've done in the rest of the code.

src/OpenSage.Game/Data/Apt/Characters/Character.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                //must be the root object. Movie does contain itself so, do a simple check
+                case CharacterType.Movie:
+                    if (container.IsEmpty)
+                    {
+                        container.IsEmpty = false;
+                        ch = Movie.Parse(reader, container);
+                    }
+                    else
+                        return null;
```
tgjones: A tiny thing, but please could add braces around this `else` branch too?

src/OpenSage.Game/Data/Utilities/Extensions/BinaryUtility.cs
```diff
@@ -15,4 +16,12 @@ static BinaryUtility()
         /// </summary>
         public static Encoding AnsiEncoding { get; }
     }
+
+    internal abstract class BinaryObject
```
tgjones: Is this used anywhere?

src/OpenSage.Game/Data/Apt/AptFile.cs
feliwir: In theory that can happen, but parsing doesn't really require a lot of time. My idea was to have FromFileSystemEntry load an entire finished .apt, so you can't possibly end up with a crippled AptFile

feliwir:
Did fix the mentioned issues

src/OpenSage.Game/Data/Apt/AptFile.cs
tgjones: I like the goal of not being able to have an incomplete AptFile. Let's leave it as it is, but perhaps in the future we can pass `importDict ` into `AptFile.FromFileSystemEntry`, so they can be cached globally.

tgjones:
Thank you! I did a "squash" merge, to combine the commits into a single commit, so you'll want to get your branch in sync with upstream before carrying on :)

PR #14, opened 2017-12-25
Fix controls with null header template

paavohuhtala:
(Based on #13)

Treat UI elements with null HeaderTemplate like ones set to "[NONE]"

Fixes crashes when viewing certain UI controls in the data viewer, such as the MOTD.


paavohuhtala:
There are, obviously, alternative ways of implementing this. One way would be to handle this case directly in the parser, where "[NONE]" would be coerced to null, and we'd only check for null.

tgjones:
Thank you!

PR #15, opened 2017-12-25
Add support for special registry key for Zero Hour, refactor installation location

paavohuhtala:
When installing an Origin copy of the Generals + Zero Hour bundle on my secondary workstation I found out that for some reason it uses a different registry key than my other computer for ZH. No idea why.

Since the alternative registry key requires some special handling, I implemented it and decided to move Zero Hour installation resolution to a different method entirely. As a bonus, a Generals installation is only resolved once for all possible Zero Hour installations.

### DIFF
--- a/src/OpenSage.Game/Data/InstallationLocator.cs
+++ b/src/OpenSage.Game/Data/InstallationLocator.cs
@@ -36,7 +36,7 @@ namespace OpenSage.Data
             }
         }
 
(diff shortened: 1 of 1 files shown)


tgjones:
Nice, thank you. As you'll have spotted I did some special-case work on resolving ZH installations, but I missed the different registry key for Origin installations.

PR #16, opened 2017-12-25
Skip tests for games you don't have installed

paavohuhtala:
(Based on #15)

This PR implements a custom XUnit test discoverer, which filters the tests by the games you have installed. Tests that require a specific game can be tagged with `[GameFact(SageGame.GameName)]`. I haven't classified every test yet, but I can do that if required.

The motivation behind this is that it doesn't make much sense to run tests that couldn't possibly pass because of missing data files. It also documents in a neat, declarative way which tests require which games.


tgjones:
This is a neat idea - thanks! I'm going to modify it a bit to allow multiple games to be passed to `GameFact(...)`, since (for example) the `.apt` tests should be run for both BFME I and II. But I can do that after merging.

PR #23, opened 2017-12-30
Fix map object z positioning

paavohuhtala:
Fixes map object Z positioning by _adding_ the height map value at current position to the object's Z position instead of ignoring it entirely.

Before: 

image

After:

image

WorldBuilder for reference:

image


feliwir:
Looks really good to me

tgjones:
Nice, thank you!

PR #24, opened 2018-01-01
Fix IniParser.IsFloat on non-English Windows

paavohuhtala:
Fixes test `IniFileTests.CanReadIniFiles` for languages that don't use the decimal point.


tgjones:
Thanks!

PR #29, opened 2018-01-02
Map loading fixes

### LINKED ISSUE #22: Handle MapObjects being placed outside heightmap boundaries
`LoadObjects` in `MapLoader.cs` should do a bounds check before calling `GetHeight`. If the object is outside of height map boundaries, it should use a different height.

I don't know what that height should be. I encountered this in `CHI07.map`, where two jets were placed outside boundaries so that they can fly in in a cinematic. Maybe `PreferredHeight` for things that have a locomotor, and 0 for others?

### LINKED ISSUE #27: Use waypoint path labels as path identifiers
image

After fixing #22 `CHI07.map` still crashes the map loader, because waypoint path `INTROConveyMove` doesn't exist. I looked at the map loader and apparently OpenSAGE uses the name of the first node as the name of the waypoint path:

```csharp
foreach (var waypointPath in mapFile.WaypointsList.WaypointPaths)
{
    var start = result.Settings.Waypoints[waypointPath.StartWaypointID];
    var end = result.Settings.Waypoints[waypointPath.EndWaypointID];

    result.Settings.WaypointPaths[start.Name] = new Settings.WaypointPath(
        start, end);
}
```

WorldBuilder and the scripting system seem to identify paths by their _waypoint path label_. You can configure up to 3 labels per path, and any of them is a valid identifier for the path.

image

From what I've gathered, these are not currently parsed.

paavohuhtala:
Fixes #22, fixes #27.

Fixes a number of campaign maps. Every map in the China campaign loads except for `CHI03`.

* Check if the coordinate is within bounds in `HeightMap.GetHeight`, return 0 if not.
* Parse `wayPointPathLabel` properties and store them in `Settings.Waypoint`s.
* Implement `WaypointPathCollection` which allows looking up `WaypointPath`s by their first node and their label.
* Fix `MoveCameraAlongWaypointPathAction`.
* As a bonus, add two self-explanatory LINQ extensions: `WhereNot` and `ToSet`. Not absolutely necessary, but I though they'd be useful.

### DIFF
--- a/src/OpenSage.Game/Data/InstallationLocator.cs
+++ b/src/OpenSage.Game/Data/InstallationLocator.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using OpenSage.Utilities.Extensions;
 
 namespace OpenSage.Data
 {
@@ -111,7 +112,7 @@ namespace OpenSage.Data
         private IEnumerable<string> GetValidPaths(IEnumerable<string> paths)
         {
             return paths
-                .Where(p => !string.IsNullOrWhiteSpace(p))
+                .WhereNot(string.IsNullOrWhiteSpace)
                 .Distinct()
                 .Where(Directory.Exists);
         }
--- a/src/OpenSage.Game.Tests/GameTestDiscoverer.cs
+++ b/src/OpenSage.Game.Tests/GameTestDiscoverer.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
 using OpenSage.Data;
+using OpenSage.Utilities.Extensions;
 using Xunit.Abstractions;
 using Xunit.Sdk;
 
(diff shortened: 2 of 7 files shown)


src/OpenSage.Game/Data/InstallationLocator.cs
```diff
@@ -111,7 +112,7 @@ private static string GetRegistryValue(string keyName, string valueName)
         private IEnumerable<string> GetValidPaths(IEnumerable<string> paths)
         {
             return paths
-                .Where(p => !string.IsNullOrWhiteSpace(p))
+                .WhereNot(string.IsNullOrWhiteSpace)
```
tgjones: It's subjective, but to me this isn't as clear as the original code - probably just because I'm very used to seeing `Where` and I'm not used to seeing `WhereNot`.

tgjones:
This is fantastic, thank you!

src/OpenSage.Game/Data/InstallationLocator.cs
paavohuhtala: Yeah, it's definitely subjective. I prefer avoiding lambdas (which is pretty difficult in C#) when not necessary, and in my opinion `!` is easier to miss than `not`.

There are options, of course. I did consider using the name `Reject` instead, as it's used in many JS FP libraries such as Lodash and Ramda. LISPs including Clojure and Emacs LISP call it`remove`.

PR #35, opened 2018-01-03
Add a project for the game executable

paavohuhtala:
Adds a new project for the game executable. It's Windows-only for now, so I named it `OpenSAGE.Windows`. It defaults to Generals and loads the main menu and the background scene.

The only change to the core engine is the addition of a `HideImage` property to `UIElement`, which is used to hide the background image in the main menu.

There's also an another patch coming related to `WaypointPath`s, which I used for the [Twitter video](https://twitter.com/paavohtl/status/948352514814038016).


src/OpenSage.Game/Gui/Wnd/Elements/UIElement.cs
```diff
@@ -145,6 +145,8 @@ public bool Highlighted
             }
         }
 
+        public bool HideImage { get; set; }
```
tgjones: I'm pretty sure this is actually handled by the `W3DNoDraw` callback, as set in `MainMenu.wnd`, line 18. The only reason I didn't implement that yet is because it was nice to see the backdrop image - but now that you're starting to add the actual shell map, we can go ahead and implement `W3DNoDraw` in the `WndCallbacks` class.

tgjones:
I think the launcher project should be named OpenSage.Launcher.Windows”, to make its purpose clearer. Would you mind making that change? Then I think this is good to be merged.

paavohuhtala:
@tgjones Renamed.

tgjones:
Thanks!

PR #38, opened 2018-01-05
Updated Nuget packages for Xunit and the Test SDK

feliwir:
This change is necessary, because the tests aren't properly detected or run, in some instances (including my machine). By updating the NuGet Packages everything is working again as expected


tgjones:
Thank you!

PR #43, opened 2018-01-17
Add IGameDefinition, GameFactory and mod DLLs

paavohuhtala:
* Add a DLL project for every supported game. At least for now expansion packs and base games share the same DLL, but that can be changed in the future.
* Add `IGameDefinition`. Game definitions are singleton objects used for configuring the engine for the target game. They also contain unchanging metadata like the display name and the registry keys used for detecting the game.
* Add `GameFactory`, which creates `Game` instances from a `IGameDefinition` and `IInstallationLocator` or from a `GameInstallation`.
* Massively simplify `GameInstallation`.
  * All hard-coded information was moved to `GameDefinition`s.
  * `GameInstallation` now references the `GameDefinition` it's for, and supports base games.
* Remove ZH specific hacks from `RegistryInstallationLocator`.
* Move `WndCallbacks` to `SageGame.Mods.Generals`.
* Rename `CallbackUtility` to `WndCallbackResolver`, and make it non-static.


src/OpenSage.DataViewer/Framework/InstallationUtility.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using OpenSage.Data;
+using OpenSage.Mods.BuiltIn;
+
+namespace OpenSage.DataViewer.Framework
+{
+    // TODO: Some of this could be moved to the engine.
```
paavohuhtala: This comment is outdated - there used to be more stuff here.

tgjones:
Thank you! This sets us up very well for the next stage of gameplay development.

PR #52, opened 2018-01-19
Add parsing of all Instructions used in BFME II MainMenu

feliwir:
This PR adds parsing for most of the Instructions, The instructions not yet implemented are used not much at all. Next step is to add the VM executing the instructions


src/OpenSage.Game/Data/Utilities/Extensions/BinaryReaderExtensions.cs
```diff
@@ -275,6 +277,30 @@ public static IndexedTriangle ReadIndexedTri(this BinaryReader reader)
                 reader.ReadUInt16());
         }
 
+        public static ButtonRecord ReadButtonRecord(this BinaryReader reader)
```
tgjones: This method, and `ReadButtonAction`, should be static `Parse` methods on `ButtonRecord` and `ButtonAction` respectively, because they're specific to .apt.

src/OpenSage.Game/Gui/Apt/ActionScript/Opcodes/Basic.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace OpenSage.Gui.Apt.ActionScript.Opcodes
+{
+    //declare a pool of constants that will be used in the following code
```
tgjones: In general, for this type of comment on classes and methods, you can use doc comments, so that they'll show up in IntelliSense:

``` csharp
/// <summary>
/// Declare a pool...
/// </summary>
public sealed class ConstantPool
```

src/OpenSage.Game/Gui/Apt/ActionScript/Opcodes/Instruction.cs
```diff
@@ -0,0 +1,214 @@
...
+    {
+        public abstract InstructionType Type { get; }
+        //the size in bytes for this instruction (not including the opcode size)
+        public virtual uint Size { get { return 0; } set { } }
+        public virtual List<Value> Parameters { get; set; }
+
+        public abstract void Execute();
+    }
+
+    public sealed class Padding : InstructionBase
```
tgjones: Why does this need to be an instruction subclass?

Issue #44, opened 2018-01-19
Use cross-platform text rendering library


tgjones:
Switch from Direct2D to SixLabors.ImageSharp.Drawing for drawing text.

PR #53, opened 2018-01-24
Replace Direct2D with custom shaders (for 2D drawing) and ImageSharp.Drawing (for text rendering)

tgjones:


tgjones:
This fixes #44.

PR #56, opened 2018-02-01
Collision components

paavohuhtala:
* Rewrite ray-AABB collision
  * The behaviour is not exactly the same as the current implementation, but it's much simpler and faster. If there are issues we can always rollback to the previous version.
* Implement abstract `ColliderComponent`
  * Implement `BoxColliderComponent`, `SphereColliderComponent` and `CylinderColliderComponent`
  * `BoxColliderComponent` implements ray-OBB collision by transforming the ray into object space. 
  * `CylinderColliderComponent` uses a bounding box for the implementation, for now.
* Add `Transform` method to `Ray`
* Implement `DebugEntityPickerSystem` and the associated input handler
  * This is essentially a prototype for unit selection and a collider cache. A dedicated collider cache could eventually use an octree, but for now a list is enough.
  * Clicking on objects scales them down by 50% - this was the simplest way to highlight the feature.


src/OpenSage.Game/Mathematics/Ray.cs
```diff
@@ -122,6 +98,11 @@ public Ray(Vector3 position, Vector3 direction)
             return result;
         }
 
+        public Ray Transform(Matrix4x4 world)
```
tgjones: This should have an `in` modifier to avoid copying the (large) `Matrix4x4` struct.

tgjones:
This is really great, thank you :)

PR #58, opened 2018-02-04
Apt work

feliwir:
- Parse PlaceObjectActions (action that a triggered under certain conditions)
- Add new instructions
- Put builtin ActionScript functions/objects into a seperate class

### DIFF
--- a/src/OpenSage.Game/Data/Apt/Characters/Button.cs
+++ b/src/OpenSage.Game/Data/Apt/Characters/Button.cs
@@ -4,15 +4,15 @@ using System.IO;
 using System.Numerics;
 using OpenSage.Data.Utilities.Extensions;
 using OpenSage.Gui.Apt.ActionScript;
-using OpenSage.Gui.Apt.ActionScript.Opcodes;
 using OpenSage.Mathematics;
 
 namespace OpenSage.Data.Apt.Characters
 {
 
     [Flags]
-    public enum ButtonRecordFlags : uint
+    public enum ButtonRecordFlags : byte
     {
+        None = 0,
         StateUp = 1,
         StateOver = 2,
         StateDown = 4,
@@ -29,7 +29,7 @@ namespace OpenSage.Data.Apt.Characters
         public ColorRgba Color;
         public Vector4 Unknown;
 
-        public ButtonRecord(ButtonRecordFlags flags, uint character, int depth, Matrix2x2 rotscale,
+        public ButtonRecord(ButtonRecordFlags flags,uint reserved, uint character, int depth, Matrix2x2 rotscale,
             Vector2 translation, ColorRgba color, Vector4 unknown)
         {
             Flags = flags;
@@ -44,7 +44,8 @@ namespace OpenSage.Data.Apt.Characters
         public static ButtonRecord Parse(BinaryReader reader)
         {
             return new ButtonRecord(
-                reader.ReadUInt32AsEnumFlags<ButtonRecordFlags>(),
+                reader.ReadByteAsEnumFlags<ButtonRecordFlags>(),
+                reader.ReadUInt24(),
                 reader.ReadUInt32(),
                 reader.ReadInt32(),
                 reader.ReadMatrix2x2(),
@@ -123,7 +124,7 @@ namespace OpenSage.Data.Apt.Characters
         {
             var button = new Button();
 
-            button.IsMenu = reader.ReadBooleanUInt32Checked();
+            button.IsMenu = Convert.ToBoolean(reader.ReadUInt32());
             button.Bounds = reader.ReadVector4();
             var tc = reader.ReadUInt32();
             var vc = reader.ReadUInt32();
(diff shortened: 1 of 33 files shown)


src/OpenSage.Game/Data/Apt/Characters/Button.cs
```diff
@@ -44,7 +44,8 @@ public struct ButtonRecord
         public static ButtonRecord Parse(BinaryReader reader)
         {
             return new ButtonRecord(
-                reader.ReadUInt32AsEnumFlags<ButtonRecordFlags>(),
+                reader.ReadByteAsEnumFlags<ButtonRecordFlags>(),
```
tgjones: In the binary file, is the enum actually stored in a uint32? If so, I think this was better before this change.

src/OpenSage.Game/Data/Apt/Characters/Button.cs
```diff
@@ -123,7 +124,7 @@ public static Button Parse(BinaryReader reader)
         {
             var button = new Button();
 
-            button.IsMenu = reader.ReadBooleanUInt32Checked();
+            button.IsMenu = Convert.ToBoolean(reader.ReadUInt32());
```
tgjones: What's the reason for this change?

src/OpenSage.Game/Data/Utilities/Extensions/BinaryReaderExtensions.cs
```diff
@@ -396,6 +404,28 @@ public static List<T> ReadListAtOffset<T>(this BinaryReader reader, Func<T> crea
             return result;
         }
 
+        public static List<T> ReadFixedSizeListAtOffset<T>(this BinaryReader reader, Func<T> creator, uint size) where T : class
```
tgjones: What's the reason for this, in addition to `ReadFixedSizeArrayAtOffset`?

src/OpenSage.Game/Graphics/SpriteBatch.cs
```diff
@@ -73,7 +73,7 @@ public SpriteBatch(ContentManager contentManager)
 
         private ref SpriteBatchItem CreateBatchItem()
         {
-            if (_currentBatchIndex > _batchItems.Length)
+            if (_currentBatchIndex >= _batchItems.Length)
```
tgjones: Thanks :)

src/OpenSage.Game/Gui/Apt/ActionScript/Function.cs
```diff
@@ -1,13 +1,36 @@
...
+﻿using System;
+using System.Collections.Generic;
 using OpenSage.Gui.Apt.ActionScript.Opcodes;
 
 namespace OpenSage.Gui.Apt.ActionScript
 {
+    [Flags]
+    public enum FunctionPreloadFlags
+    {
+        PreloadExtern = 0x010000,   //this seems to be added by EA
```
tgjones: It's not critical, but I usually like enum values to be in ascending numeric order, unless there's a good reason not to.

src/OpenSage.Game/Gui/Apt/ActionScript/Opcodes/Object.cs
```diff
@@ -100,18 +117,44 @@ public override void Execute(ActionContext context)
...
     public sealed class GetProperty : InstructionBase
     {
         public override InstructionType Type => InstructionType.GetProperty;
 
         public override void Execute(ActionContext context)
         {
-            throw new NotImplementedException();
+            var property = context.Stack.Pop();
+            var target = context.GetTarget(context.Stack.Pop().ToString());
+            var type = (PropertyType) property.ToInteger();
```
tgjones: If this is a common thing to do, would it be worth having a `property.ToEnum<TEnum>()` method?

src/OpenSage.Game/Data/Apt/Characters/Button.cs
feliwir: Couldn't do it like before, because i encountered some other bits in the uint24. I don't know what those mean yet, but it makes the Enum verification fail

src/OpenSage.Game/Data/Apt/Characters/Button.cs
feliwir: I have no idea, why i did that :)

src/OpenSage.Game/Data/Utilities/Extensions/BinaryReaderExtensions.cs
feliwir: I needed it for the new DefineFunction2. I guess i should decide wether i use Array or List at some point

src/OpenSage.Game/Gui/Apt/ActionScript/Function.cs
feliwir: I did that because that's the order of bits. Otherwise i always get confused

src/OpenSage.Game/Gui/Apt/ActionScript/Opcodes/Object.cs
feliwir: property is of type Value, where i could add that. But the Value type has lots of room for improvements anyways ;D

tgjones:
Great work, thank you :)

PR #61, opened 2018-02-08
Optimize particle systems

paavohuhtala:
* Change `ParticleAlphaKeyframe` and `ParticleColorKeyframe` into structs to reduce allocations during rendering.
* Reuse alpha keyframe lists
* Extract the logic for finding the current and previous keyframes into a separate function
* Reduce large struct copies by adding `in` to `Vector3` arguments and passing world matrices by a readonly reference.
  * Note that Vector3 copies can't be completely eliminated, because `System.Numerics` doesn't use `in` parameters yet.
* Fix engine crash when trying to initialize particle systems with burst count of 0. Fixed by adding an early return to `ParticleSystem` constructor, and null checks in `Update` and `BuildRenderList`.

With this patch particle systems should allocate **0 bytes** after initialisation.

### DIFF
--- a/src/OpenSage.Game/Graphics/ParticleSystems/ParticleSystem.cs
+++ b/src/OpenSage.Game/Graphics/ParticleSystems/ParticleSystem.cs
@@ -17,10 +17,11 @@ using Veldrid;
 
 namespace OpenSage.Graphics.ParticleSystems
 {
-    public delegate ref readonly Matrix4x4 GetMatrixReferenceDelegate();
 
     public sealed class ParticleSystem : DisposableBase
     {
+        public delegate ref readonly Matrix4x4 GetMatrixReferenceDelegate();
+
         private readonly GetMatrixReferenceDelegate _getWorldMatrix;
 
         private readonly GraphicsDevice _graphicsDevice;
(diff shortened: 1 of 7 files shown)


src/OpenSage.Game/Graphics/ParticleSystems/ParticleSystem.cs
```diff
@@ -432,6 +403,26 @@ private void UpdateParticle(ref Particle particle)
             particle.Timer += 1;
         }
 
+        private static void FindKeyframes<T>(int timer,
+            IReadOnlyList<T> keyFrames,
+            out T next, out T prev)
```
paavohuhtala: This could return a tuple instead of having two out params, but this resulted in cleaner code. Performance should be identical.

src/OpenSage.Game/Graphics/ParticleSystems/ParticleSystem.cs
```diff
@@ -432,6 +403,26 @@ private void UpdateParticle(ref Particle particle)
...
+            where T : struct, IParticleKeyframe
```
paavohuhtala: This function is generic to avoid boxing the results.

src/OpenSage.Game/Graphics/ParticleSystems/ParticleSystem.cs
```diff
@@ -360,29 +376,12 @@ private void UpdateParticle(ref Particle particle)
...
-                nextC = prevC;
-            }
-            if (prevC != nextC)
+            FindKeyframes(particle.Timer, _colorKeyframes, out var nextC, out var prevC);
+
+            if (!prevC.Equals(nextC))
             {
                 var colorInterpoland = (float) (particle.Timer - prevC.Time) / (nextC.Time - prevC.Time);
-                particle.Color = Vector3Utility.Lerp(ref prevC.Color, ref nextC.Color, colorInterpoland);
+                particle.Color = Vector3Utility.Lerp(in prevC.Color, in nextC.Color, colorInterpoland);
```
tgjones: I haven't been using the `in` keyword in call sites - it's optional, and I think not including it better matches the default behaviour for reference types.

src/OpenSage.Game/Graphics/ParticleSystems/ParticleSystem.cs
tgjones: That's interesting - are you sure it doesn't box when it calls the `Time` property? Since it knows about that property through an interface, I wasn't sure if it would box the struct, even though because of the generic type it could access it directly.

tgjones:
This is really good stuff, thank you! It will be good to apply these sorts of techniques as we progress with more engine code.

PR #62, opened 2018-02-09
Implement file export in data viewer

paavohuhtala:


tgjones:
Thank you! I don't want to get too sidetracked by editor-type features, but this seems like a quick win.

PR #64, opened 2018-02-19
Add basic launch parameters to launcher using System.CommandLine

paavohuhtala:
* `--noshellmap`
  * Skips loading the shell map, speeding up initial loading time.
  * Present in SAGE, but with only one dash.
  * Related to this: draw the static background image in Generals & ZH main menu if there's no 3D scene.
* `--game`
  * Chooses which game to start. Generals is still the default.
  * Currently takes the game name as a case-insensitive `SageGame` entry. I'd like to replace this with a shorter string identifier (added to `IGameDefinition`), which could be used for other things such as configuration files, debug information and maybe even multiplayer lobbies.
* Add `IMainMenuSource`, `WndMainMenuSource` and `AptMainMenuSource`
* Add `Configuration` type. The only implemented setting right now is `bool LoadShellMap`.
* Add a reference to `IGameDefinition` in `Game`
* Move shell map handling to `Game`

### DIFF
--- a/src/OpenSage.Launcher/OpenSage.Launcher.csproj
+++ b/src/OpenSage.Launcher/OpenSage.Launcher.csproj
@@ -15,7 +15,7 @@
   </ItemGroup>
   
   <ItemGroup>
-    <PackageReference Include="System.CommandLine" Version="0.1.0-preview2-180210-2" />
+    <PackageReference Include="System.CommandLine" Version="0.1.0-preview2-180220-2" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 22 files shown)


src/OpenSage.Launcher/OpenSage.Launcher.csproj
```diff
@@ -13,6 +13,10 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="System.CommandLine" Version="0.1.0-preview2-180220-2" />
```
tgjones: Looks like this is from MyGet, not NuGet? So you'll need to add the [dotnet-corefxlab MyGet feed](https://dotnet.myget.org/feed/dotnet-corefxlab/package/nuget/System.CommandLine) to [this file](https://github.com/OpenSAGE/OpenSAGE/blob/master/NuGet.config).

paavohuhtala: Looks like I had it in my global Nuget config, so it built on my computer but failed in CI - I copied the line to NuGet.config and now it builds.

tgjones:
Thank you!

PR #66, opened 2018-02-25
Work on an audiosystem

feliwir:
- adds .wav loading
- add a .wav viewer that does playback the sound in a loop
still WIP (sounds don't stop when switching the view)

### DIFF
--- a/src/OpenSage.DataViewer/UI/Viewers/WavView.cs
+++ b/src/OpenSage.DataViewer/UI/Viewers/WavView.cs
@@ -12,7 +12,6 @@ namespace OpenSage.DataViewer.UI.Viewers
     class WavView : GameControl
     {
         WavFile _file;
-        AudioSource _source;
         Game _game;
 
         public WavView(FileSystemEntry entry, Func<IntPtr, Game> createGame)
@@ -21,18 +20,12 @@ namespace OpenSage.DataViewer.UI.Viewers
             CreateGame = h =>
             {
                 _game = createGame(h);
-
-                var wavFile = _game.ContentManager.Load<AudioBuffer>(entry.FilePath);
-                var source = _game.Audio.AddSource(wavFile, true);
+            
+                var source = _game.Audio.PlayFile(entry.FilePath, true);
                 source.Play();
 
                 return _game;
             };         
-        }
-
-        ~WavView()
-        {
-            _game.Audio.RemoveSource(_source);
         }
     }
 }
--- a/src/OpenSage.DataViewer/UI/Viewers/WavView.cs
+++ b/src/OpenSage.DataViewer/UI/Viewers/WavView.cs
@@ -2,18 +2,37 @@
 using System.Collections.Generic;
 using System.Text;
 using Eto.Forms;
+using OpenSage.Audio;
 using OpenSage.Data;
 using OpenSage.Data.Wav;
+using OpenSage.DataViewer.Controls;
 
 namespace OpenSage.DataViewer.UI.Viewers
 {
-    class WavView : Panel
+    class WavView : GameControl
     {
         WavFile _file;
+        AudioSource _source;
+        Game _game;
 
-        public WavView(FileSystemEntry entry)
+        public WavView(FileSystemEntry entry, Func<IntPtr, Game> createGame)
         {
             _file = WavFile.FromFileSystemEntry(entry);
+            CreateGame = h =>
+            {
+                _game = createGame(h);
+
+                var wavFile = _game.ContentManager.Load<AudioBuffer>(entry.FilePath);
+                var source = _game.Audio.AddSource(wavFile, true);
+                source.Play();
+
+                return _game;
+            };         
+        }
+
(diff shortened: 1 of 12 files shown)


src/OpenSage.DataViewer/UI/Viewers/WavView.cs
```diff
@@ -0,0 +1,38 @@
...
+
+                var wavFile = _game.ContentManager.Load<AudioBuffer>(entry.FilePath);
+                var source = _game.Audio.AddSource(wavFile, true);
+                source.Play();
+
+                return _game;
+            };         
+        }
+
+        ~WavView()
```
tgjones: This should override `OnUnload`.

src/OpenSage.Game/Audio/AudioBuffer.cs
```diff
@@ -0,0 +1,106 @@
...
+                    }
+                    break;
+            }
+
+
+            AL10.alBufferData(_handle, format, wavFile.Buffer, wavFile.Size, wavFile.Fequency);
+            AudioSystem.alCheckError();
+        }
+
+        public TimeSpan Duration
```
tgjones: You mentioned this can't be calculated upfront in the constructor because you want to re-use this class for streamed audio. If so, I think it would be better to make this a `GetDuration()` method. .NET guidelines suggest using methods instead of properties for "expensive" computations like this one.

src/OpenSage.Game/Audio/AudioBuffer.cs
```diff
@@ -0,0 +1,106 @@
...
+                for(int i=0;i<_usageList.Count;++i)
+                {
+                    _usageList[i].Dispose();
+                }
+
+                AL10.alDeleteBuffers(1, ref _handle);
+                AudioSystem.alCheckError();
+                _handle = 0;
+                
+                base.Dispose(disposeManagedResources);
```
tgjones: The call to `base.Dispose` should be outside of the `if`.

src/OpenSage.Game/Audio/AudioBuffer.cs
```diff
@@ -0,0 +1,106 @@
...
+using OpenSage.Content;
+using OpenSage.Data.Wav;
+using OpenAL;
+
+namespace OpenSage.Audio
+{
+    public sealed class AudioBuffer : DisposableBase
+    {
+        private uint _handle;
+        private List<AudioSource> _usageList;
```
tgjones: I don't think this usage list is necessary, assuming that `AudioBuffer` instances always live longer than the `AudioSource` that use them, as I think they should.

src/OpenSage.Game/Audio/AudioSource.cs
```diff
@@ -0,0 +1,56 @@
...
+            _duration = buffer.Duration;
+        }
+
+        protected override void Dispose(bool disposeManagedResources)
+        {
+            if (_handle > 0)
+            {
+                Stop();
+                AL10.alDeleteSources(1, ref _handle);
+                base.Dispose(disposeManagedResources);
```
tgjones: The call to `base.Dispose` should be outside of the `if`.

src/OpenSage.Game/Audio/AudioSystem.cs
```diff
@@ -0,0 +1,78 @@
...
+            _context = ALC10.alcCreateContext(_device, null);
+            alcCheckError();
+
+            ALC10.alcMakeContextCurrent(_context);
+            alcCheckError();
+
+            _sources = new List<AudioSource>();
+        }
+
+        ~AudioSystem()
```
tgjones: This should be an override of `Dispose`.

src/OpenSage.Game/Audio/AudioSystem.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            ALC10.alcMakeContextCurrent(IntPtr.Zero);
+            ALC10.alcDestroyContext(_context);
+            ALC10.alcCloseDevice(_device);
+        }
+
+        public AudioSource AddSource(AudioBuffer buffer,bool loop=false)
```
tgjones: I think this API could be `public AudioSource PlayFile(string fileName, bool loop = false)`. We can reserve the `Play` method name for later, when we'll want to be able to play `AudioEvent`s by name.

src/OpenSage.Game/Audio/AudioSystem.cs
```diff
@@ -0,0 +1,75 @@
...
+            alcCheckError();
+
+            _sources = new List<AudioSource>();
+        }
+
+        protected override void Dispose(bool disposeManagedResources)
+        {
+            ALC10.alcMakeContextCurrent(IntPtr.Zero);
+            ALC10.alcDestroyContext(_context);
+            ALC10.alcCloseDevice(_device);
```
tgjones: Could you add a call to `base.Dispose` here?

tgjones:
Thank you!

PR #67, opened 2018-02-26
Environment Variable Installation Locator

Nostritius:
This PR should implement a second Installation Locator using environment variables, mainly to support none-windows systems.

### DIFF
--- a/src/OpenSage.DataViewer/Framework/InstallationUtility.cs
+++ b/src/OpenSage.DataViewer/Framework/InstallationUtility.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using OpenSage.Data;
 using OpenSage.Mods.BuiltIn;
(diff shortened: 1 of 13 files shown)


Nostritius:
@feliwir I have changed this

paavohuhtala:
@Nostritius This is good stuff!

One change I'd like to see to this PR is that instead of having a separate property for the environment variable in `IGameDefinition` , we'd have a short game-specific string identifier instead (`string Identifier { get; }`). The environment variable name would then be automatically derived from this identifier, by e.g uppercasing it and appending `_PATH` to it.

This identifier was brought up in #64 and it would be useful for lots of other uses cases, including command line argument and configuration file installation locators, for the `--game` parameter, mod specific configuration and so on.

I think the identifiers should be as short as possible while remaining recognisable without a manual, like you've already done with the environment variable names. They should only use filesystem and command line compatible characters and not use any whitespace. I leave the specific details to you. 😄

Nostritius:
@paavohuhtala I have now added a generic identifier to the definitions and used it to generate the Environment variable from it.

tgjones:
This is great stuff - thanks for your contribution @Nostritius.

PR #68, opened 2018-02-26
Use a regex to get localized launcher images

Nostritius:
This PR changed the launcher image loading to a regex to make it possible to load launcher images of different languages,  because for example in Bfme they are called Germansplash.jpg in the german version (which I use).


Qibbi:
If you're using registry/env variables I would suggest loading the language from there (as the game does, it just appends it) and if it's not present use either a default or the suggested wildcard method.

tgjones:
I think @Qibbi's comment is the right approach here. We need to know the language anyway in order to load the correct translation file. Again, we could guess it based on what is present in the file system, but since that information is in the registry, we should use it.

We might also need to extend your work in #67 to allow supplying the language through an environment variable, but we can deal with that separately.

PR #69, opened 2018-02-26
Catch exceptions of corrupted big files

Nostritius:
I recognized that OpenSAGE crashed with my battle for middle earth installation, because of a corrupted big file which the installer placed. I added some lines to skip over such files.

### DIFF
--- a/src/OpenSage.Game/Data/FileSystem.cs
+++ b/src/OpenSage.Game/Data/FileSystem.cs
@@ -34,7 +34,15 @@ namespace OpenSage.Data
                 var ext = Path.GetExtension(file).ToLower();
                 if (ext == ".big")
                 {
-                    var archive = new BigArchive(file);
+                    BigArchive archive;
+                    try
+                    {
+                        archive = new BigArchive(file);
+                    }
+                    catch (Exception e)
+                    {
+                        continue;
+                    }
 
                     _bigArchives.Add(archive);
 


Qibbi:
If you're using registry/env variables I would suggest loading the language from there (as the game does, it just appends it) and if it's not present use either a default or the suggested wildcard method.

Qibbi:
TW would craah iirc (at least if you're not using try-add-big which encapsulates it in a try block. Dunno about earlier games.

tgjones:
@Nostritius did you see my questions above? I'm curious about why your BFME installation has a corrupt `.big` file, and whether the original game can handle it.

Nostritius:
First sorry, that i have not yet answered. I was pretty busy and it was more difficult than i expected to get it running on windows (I have just installed but not used the bfme to test it against OpenSAGE. The program definitly ignores the EnglishAudio.big and starts even with this file being in the directory and in the filelist.txt).

tgjones:
Which locale is it trying to load? If it’s not English, I wonder if EnglishAudio.big is skipped altogether, rather than and attempted and ignored. I’ll test that.

Nostritius:
I have loaded the german version from the battle for middle earth anthology. The EnglishAudio.big file was already contained in the corresponding language specific zip file on the installer disc. I have also viewed it in a hex editor, and it contains exactly two bytes which are 0x3F, and this corresponds to the Ascii "??" string.

tgjones:
This was fixed by https://github.com/OpenSAGE/OpenSAGE/commit/d7c2af80193a8c24c435dfba195027a5abdfca90

PR #71, opened 2018-03-01
Rename SageGame entries for BFME games, add BFME2 expansion

paavohuhtala:
Huge diff, but very few manual changes.

* Rename `SageGame` entries for BFME and BFME2
* Rename BFME II assembly (and definition, and project) to BFME2
* Add the BFME2 expansion pack:
  * Add `Bfme2Rotwk` to `SageGame`
    * Add this entry to e.g switch statements where `Bfme2` was used already
  * Add `Bfme2RotwkDefinition` to `OpenSage.Mods.Bfme2`, including the registry key.
  * Run some APT-related tests for the expansion pack
  * Add the new `ShellMapOn` property to `GameData` and parse it.

### DIFF
--- a/src/OpenSage.Game.Tests/Data/Apt/AptFileTest.cs
+++ b/src/OpenSage.Game.Tests/Data/Apt/AptFileTest.cs
@@ -14,7 +14,7 @@ namespace OpenSage.Tests.Data.Apt
             _output = output;
         }
 
-        [GameFact(SageGame.BattleForMiddleEarth, SageGame.BattleForMiddleEarthII)]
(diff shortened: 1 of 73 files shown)


tgjones:
Great stuff, thank you!

Does ROTWK use data from BFME2's folder, as ZH does with Generals?

feliwir:
It does use BFME II data

paavohuhtala:
Yes, based on two observations:

* The maps use textures from the base game, which will crash the engine if `BaseGame` is null.
* The installation folder is less than half the size of BFME2, even though it has more units, buildings etc than BFME2.

PR #73, opened 2018-03-01
[launcher] Fix soft_oal.dll not being copied to the output directory

SomeAnon42:
…and explicitly set PlatformTarget to x64 to avoid BadImageFormatException (as x86 is the implicit default for SDK-based netfx projects)


tgjones:
This is great, thank you for your contribution!

PR #75, opened 2018-03-03
[WIP] Add Player class

paavohuhtala:


src/OpenSage.Game/Logic/Player.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Collections.Generic;
+using OpenSage.Content;
+using OpenSage.Data.Ini;
+using OpenSage.Logic.Object;
+using OpenSage.Utilities.Extensions;
+
+namespace OpenSage.Logic
+{
+    // TODO: Abstract this out, or try to include all factions from every game?
+    public enum Side
```
tgjones: What benefit is there from this being an enum? I don't know if we'll need to have many/any `switch` statements based on this value.

tgjones:
Thank you!

src/OpenSage.Game/Logic/Player.cs
paavohuhtala: An enum felt like a better idea than an arbitrary string, but you're right that we're unlikely to benefit from the additional type safety.

Qibbi:
I would suggest using a POID like later games do.

paavohuhtala:
@Qibbi I'm not familiar with the term, could you tell me more? Is it like a player-specific GUID?

Qibbi:
Pipeline Object ID, in TW+ it's basically an asset reference, either just the lowercase hash of the asset name, or a hash of the type name followed by the lowercase hash of the asset name.

paavohuhtala:
@Qibbi And what is it used for? Networking, serialisation, saves?

Qibbi:
Basically in everything to reference an object.

Qibbi:
Or actually rather an object archetype. In networking for example objects on the map are just referenced by index in the global object table. (difference between pipeline objects and instantiated objects)

PR #78, opened 2018-03-06
Add StartingBuilding and StartingUnit0 to game objects

tgjones:
This is all very hard-coded, temporary code,  but it should serve as a stepping stone to something better.


tgjones:
We'll need to revisit this code at some point soon, but it will do for now.

PR #80, opened 2018-04-01
Get Launcher running on macOS

tgjones:


tgjones:
Launcher is now looking pretty good on macOS!

2018-04-11 macos launcher

PR #85, opened 2018-04-05
Add --map command to launcher to help with debugging

paavohuhtala:
I got tired of having to edit `Program.cs` every time I want to test something in-game.


tgjones:
Thanks! We may not want this here forever but it makes sense for now.

PR #92, opened 2018-04-12
Allow multiple active installation locators

tgjones:
Locators are queried in priority order to find installations of a game.


tgjones:
This means that on Windows, the registry will be searched first, falling back to environment variables.

PR #94, opened 2018-04-13
Input handler priority

paavohuhtala:
* Go through `InputMessageHandler`s in the priority order instead of registration order. This uses the `HandlingPriority` enum and the abstract `Priority` property in `InputMessageHandler`. The algorithm is implemented as an iterator: for each message, the handlers are looped through starting with the highest possible priority. If there is a handler matching the priority level, yield it. Decrease the priority threshold by one, and go through the handlers again.
* Prioritise selection system over UI when a selection box is being dragged
* Prioritise camera system over UI when the camera is being moved or rotated.
* Consume right click and middle click events when the cursor is over a control.

### DIFF
--- a/src/OpenSage.Game/Gui/Wnd/WndInputMessageHandler.cs
+++ b/src/OpenSage.Game/Gui/Wnd/WndInputMessageHandler.cs
@@ -13,6 +13,8 @@ namespace OpenSage.Gui.Wnd
 
         private readonly List<Control> _lastMouseOverControls = new List<Control>();
 
+        public override HandlingPriority Priority => HandlingPriority.UIPriority;
+
         public WndInputMessageHandler(WndWindowManager windowManager, Game game)
         {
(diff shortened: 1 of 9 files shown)


src/OpenSage.Game/Gui/Wnd/WndInputMessageHandler.cs
```diff
@@ -21,6 +24,20 @@ public WndInputMessageHandler(WndWindowManager windowManager, Game game)
 
         public override InputMessageResult HandleMessage(InputMessage message)
         {
+            bool GetControlAtPoint(in Point2D mousePosition, out Control control, out Point2D controlRelativePosition)
```
tgjones: My naming convention in other parts of the code is to use camelCase for local functions. Is PascalCase what ReSharper recommends?

tgjones:
Nice improvement, thank you!

PR #95, opened 2018-04-13
Controlbar improvements

paavohuhtala:
* Implement control bar abstraction, with `IControlBar` and `IControlBarSource`.
* Move control bar to its own classes
* Apply the control bar theme from `ControlBarScheme.ini`
* Remove ESC as a window-closing hotkey. The exact reasons are outlined in 5b05fd95b1b2d20835322d59587a95d2df7ce877, but the gist of it is that it was a hack from the start, and it's not going to work properly outside of the main menu.
* **Bonus!** Fix starting building / unit positions.

### DIFF
--- a/src/OpenSage.Game/Content/ContentManager.cs
+++ b/src/OpenSage.Game/Content/ContentManager.cs
@@ -53,7 +53,7 @@ namespace OpenSage.Content
 
         public TranslationManager TranslationManager { get; }
 
-        internal WndImageLoader WndImageLoader { get; }
+        public WndImageLoader WndImageLoader { get; }
 
         public ContentManager(
             Game game,
--- a/src/OpenSage.Mods.Generals/Gui/GeneralsControlBar.cs
+++ b/src/OpenSage.Mods.Generals/Gui/GeneralsControlBar.cs
@@ -1,13 +1,15 @@
 ﻿using OpenSage.Content;
 using OpenSage.Data.Ini;
+using OpenSage.Gui;
+using OpenSage.Gui.Wnd;
 using OpenSage.Gui.Wnd.Controls;
 using OpenSage.Gui.Wnd.Images;
 using OpenSage.Logic;
 using OpenSage.Mathematics;
 
-namespace OpenSage.Gui.Wnd
+namespace OpenSage.Mods.Generals.Gui
 {
-    public sealed class WndControlBar
+    public sealed class GeneralsControlBar : IControlBar
     {
         private enum ControlBarSize
         {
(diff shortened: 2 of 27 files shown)


src/OpenSage.Mods.Generals/Gui/GeneralsControlBar.cs
```diff
@@ -0,0 +1,285 @@
...
+                }
+            }
+
+            public override void Update(Player player, GeneralsControlBar controlBar)
+            {
+
+            }
+        }
+
+        private sealed class SelectedControlBarState : ControlBarState
```
paavohuhtala: This is probably not the final design, since all control bar states (except the default) are somehow related to selection. Those details need to be ironed out, but that can wait for the next PR.

src/OpenSage.Game/Content/ContentManager.cs
```diff
@@ -53,7 +53,7 @@ public sealed class ContentManager : DisposableBase
 
         public TranslationManager TranslationManager { get; }
 
-        internal WndImageLoader WndImageLoader { get; }
+        public WndImageLoader WndImageLoader { get; }
```
paavohuhtala: I had to make this public to call `WndImageLoader.CreateNormalImage` from Generals' DLL.

If we don't want to make this API public, we could add a method to `ContentManager` or `WndWindowManager` instead.

src/OpenSage.Mods.Generals/Gui/GeneralsControlBar.cs
```diff
@@ -0,0 +1,285 @@
...
+            public override void Update(Player player, GeneralsControlBar controlBar)
+            {
+                throw new System.NotImplementedException();
+            }
+        }
+    }
+
+    public sealed class GeneralsControlBarSource : IControlBarSource
+    {
+        public IControlBar Create(string side, ContentManager contentManager)
```
paavohuhtala: This is a long method, but it's very hard to split up.

src/OpenSage.Mods.Generals/Gui/GeneralsControlBar.cs
```diff
@@ -0,0 +1,285 @@
...
+            // TODO: Implement under attack indicator.
+            FindControl("WinUAttack").Hide();
+
+            // TODO: What is this?
+            FindControl("OnTopDraw").Hide();
+
+            var windowOrigin = controlBarWindow.Bounds.Location;
+            var schemeType = scheme.GetType();
+
+            Control ApplyBounds(string name, string coordPrefix)
```
paavohuhtala: I normally wouldn't use reflection for something that can be done safely without it, but the amount of boilerplate this saves is pretty huge. And since this is only called during initialisation, the performance difference is negligible.

PR #96, opened 2018-04-13
Added Thyme to the list of similar projects

neuromancer:


tgjones:
Thanks for your contribution!

However, if you look in the history of README.md, you will find exactly this link :) I decided to remove it in order to maintain some separation. We are not using any reverse engineered code in OpenSAGE, and I want to avoid any suggestion that the two projects are linked.

(This doesn't mean I think badly of the project, at all - on the contrary, they are smart people and I wish them well.)

neuromancer:
Fair enough.

PR #100, opened 2018-04-14
Use the Invariant version of ToUpper and ToLower (and fix Turkish locale)

paavohuhtala:
Fixes issues with Turkish locale. This was just a find-and-replace job - some calls could work just fine without these changes, but it's good to be consistent.


tgjones:
Thank you!

PR #101, opened 2018-04-15
Implement road rendering

tgjones:


tgjones:
So far I've implemented disconnected segments, with no junctions:

tgjones:
This isn't finished - it can only draw straight road segments, with no joins - but it might as well be merged as is. Straight roads look better than no roads.

PR #108, opened 2018-04-21
Create the window & the device with a single call

feliwir:
Has the benefit that window flags can be set depending on the backend

### DIFF
--- a/src/OpenSage.Game/GameWindow.cs
+++ b/src/OpenSage.Game/GameWindow.cs
@@ -14,6 +14,7 @@ namespace OpenSage
         public event EventHandler ClientSizeChanged;
 
         private readonly Sdl2Window _window;
+        private readonly GraphicsDevice _device;
 
         private readonly Queue<InputMessage> _messageQueue = new Queue<InputMessage>();
 
@@ -26,7 +27,7 @@ namespace OpenSage
             ClientSizeChanged?.Invoke(this, EventArgs.Empty);
         }
 
-        public GraphicsDevice GraphicsDevice { get; }
+        public GraphicsDevice GraphicsDevice => _device;
 
         public Rectangle ClientBounds
         {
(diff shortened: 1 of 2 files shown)


src/OpenSage.Game/GameWindow.cs
```diff
@@ -26,7 +27,7 @@ private void RaiseClientSizeChanged()
             ClientSizeChanged?.Invoke(this, EventArgs.Empty);
         }
 
-        public GraphicsDevice GraphicsDevice { get; }
+        public GraphicsDevice GraphicsDevice => _device;
```
tgjones: I think you can leave this property as it is, and use a local variable for `device` in the constructor, which you then assign to this property.

src/OpenSage.Game/GameWindow.cs
```diff
@@ -77,19 +64,37 @@ public GameWindow(string title, int x, int y, int width, int height, GraphicsBac
...
             if (preferredBackend != null)
             {
-                GraphicsDevice = AddDisposable(VeldridStartup.CreateGraphicsDevice(
-                    _window,
+                VeldridStartup.CreateWindowAndGraphicsDevice(
+                    windowCreateInfo,
                     graphicsDeviceOptions,
-                    preferredBackend.Value));
+                    preferredBackend.Value,
+                    out _window,
```
tgjones: Looks like you're missing a call to `AddDisposable` for the newly-created graphics device?

src/OpenSage.Game/GameWindow.cs
```diff
@@ -77,19 +64,37 @@ public GameWindow(string title, int x, int y, int width, int height, GraphicsBac
                 ResourceBindingModel = ResourceBindingModel.Improved
             };
 
+            var windowCreateInfo = new WindowCreateInfo(x, y, width, height, WindowState.Normal, title);
+
             if (preferredBackend != null)
```
tgjones: I've opened https://github.com/mellinoe/veldrid/pull/65 which will let us remove this `if`.

src/OpenSage.Game/GameWindow.cs
feliwir: I wasn’t sure how to make this to a disposable when it’s not a return type

tgjones: You can call `AddDisposable(_device);` as a separate statement.

PR #117, opened 2018-05-10
Update veldrid and shadergen

feliwir:
The updating is required because ShaderGen was producing invalid shaders for cultures that use ',' as float seperator


tgjones:
Thank you!

PR #119, opened 2018-05-15
Set alpha for map textures to opaque. Fixes black gaps in the terrain

feliwir:


tgjones:
Thank you for fixing this!

PR #121, opened 2018-05-16
Fix the text color for the OpenGL backends.

feliwir:
There is no BGRA in OpenGL, it always created a RGBA texture. Now use RGBA right away to get consistent text colors across all backends


tgjones:
Can we close this one now - we don't need it anymore right?

feliwir:
Yup

PR #122, opened 2018-05-21
Fix OpenGL issues

feliwir:
- Update Veldrid and Shadergen
- Fix integer conversion issues inside the shaders


src/OpenSage.Game/Graphics/Shaders/NormalMapped.cs
```diff
@@ -57,6 +57,7 @@ public struct MaterialConstantsType
             public float BumpScale;
             public float SpecularExponent;
             public uint /*bool*/ AlphaTestEnable;
+            public uint Padding;
```
tgjones: Please could this be `private readonly float _padding`, same as the padding values in other shaders, and wrapped in a `#pragma` that disables the "unused" warning?

feliwir: Sure will fix

tgjones:
Thank you!

PR #123, opened 2018-05-21
Add embedded fonts in case Arial isn't installed as a system font

feliwir:
This proposal is adding opensource Google Fonts as a fallback in case the Arial family isn't available. This is especially the case on linux devices and some users might not want to install those due to Licensing restrictions (not Opensourc etc.). Embedding those fonts into the assembly doesn't add more than 300kb and should make OpenSage more error prone. 

This is how it looks with the Arial family: 
arial

This is how it looks with the Roboto family:
roboto

Obviously there are some differences, but i like the look of both screenshots.


tgjones:
This is a nice idea, thank you!

PR #124, opened 2018-05-28
Work on online compilation

feliwir:
- Updates Veldrid to support native HLSL and Metal compilation
- Added Shaderc wrapper to support GLSL->Vulkan
- Always check first if bytecode already exists


tgjones:
Thank you!

Issue #89, opened 2018-04-08
Implement Language property in IInstallationLocator and its implementations


tgjones:
Once implemented, we should use it load the correct launcher image for the current locale.

This is a more long-term solution for #68.

PR #130, opened 2018-06-26
*language information for current installation as static field to use…

MarkusHiller:
… in all nessary methodths

https://github.com/OpenSAGE/OpenSAGE/issues/89


src/OpenSage.Game/LanguageSetting.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        //do we need a default?
+        private const string DefaultLanguage = "englisch";
+        private static string _detectedLanguage;
+
+        public static string Current => string.IsNullOrEmpty(_detectedLanguage) ? DefaultLanguage : _detectedLanguage;
+
+        public static void ReadFromRegistry(IGameDefinition gameDefinition)
+        {
+            var languageRegistryKey = gameDefinition.RegistryKeys.FirstOrDefault(i => i.ValueName == "Language");
```
paavohuhtala: As said in the other comment, `RegistryKeys` isn't really intended for this purpose.

What you should do instead is to either:

1. Just hardcode the registry key in this method (or something correspoding to it) and check if `IGameDefinition.Game` is `SageGame.CncGenerals` or `SageGame.CncGeneralsZeroHour`.
2. Add something like a `LanguageRegistryKey` to `IGameDefinition`. While that would technically be better (our goal is to have as little game-specific code in the engine as possible), it requires changing every single game definition and I'm not certain if there's an equivalent registry key for all (or even most) of the games.

src/OpenSage.Game/LanguageSetting.cs
```diff
@@ -0,0 +1,25 @@
...
+using OpenSage.Data;
+using OpenSage.Utilities;
+
+namespace OpenSage
+{
+    public static class LanguageSetting
+    {
+        //do we need a default?
+        private const string DefaultLanguage = "englisch";
+        private static string _detectedLanguage;
```
paavohuhtala: I don't think we should use a static field for this...

Looking up the language setting in e.g `ContentManager`'s constructor would probably make more sense. The language is only used twice, in `TranslationManager` and `WindowLoader`, so we don't need to store it for later. You can move `ReadFromRegistry` there and rename it.

src/OpenSage.Game/LanguageSetting.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Linq;
+using OpenSage.Data;
+using OpenSage.Utilities;
+
+namespace OpenSage
+{
+    public static class LanguageSetting
+    {
+        //do we need a default?
+        private const string DefaultLanguage = "englisch";
```
paavohuhtala: `"englisch"` 🙂

src/OpenSage.Game/LanguageSetting.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using OpenSage.Data;
+using OpenSage.Utilities;
+
+namespace OpenSage
+{
+    public static class LanguageSetting
+    {
+        //do we need a default?
```
paavohuhtala: Yes we do, because the registry is only available on Windows.

src/OpenSage.Game/LanguageSetting.cs
```diff
@@ -0,0 +1,25 @@
...
+        private const string DefaultLanguage = "englisch";
+        private static string _detectedLanguage;
+
+        public static string Current => string.IsNullOrEmpty(_detectedLanguage) ? DefaultLanguage : _detectedLanguage;
+
+        public static void ReadFromRegistry(IGameDefinition gameDefinition)
```
paavohuhtala: You should add some additional logic to check if the current platform (`Environment.OSVersion`) is Windows or not. AFAIK calling any of the registry methods on Mac and Linux will throw `PlatformNotSupportedException`. If the game is not running on Windows, just default to english until we have a configuration option for this.

src/OpenSage.Launcher/Program.cs
```diff
@@ -68,6 +68,7 @@ public static void Main(string[] args)
             }
 
             Platform.Start();
+            LanguageSetting.ReadFromRegistry(definition);
```
paavohuhtala: This only works for the launcher, but the engine can also be run via unit tests and the data viewer.

src/OpenSage.Game/Content/TranslationManager.cs
```diff
@@ -16,7 +16,7 @@ public TranslationManager(FileSystem fileSystem,SageGame game)
             {
                 case SageGame.CncGenerals:
                 case SageGame.CncGeneralsZeroHour:
-                    csfEntry = fileSystem.GetFile(@"Data\English\generals.csf");
+                    csfEntry = fileSystem.GetFile($@"Data\{LanguageSetting.Current}\generals.csf");
```
paavohuhtala: Take this as an optional constructor argument.

src/OpenSage.Game/Content/WindowLoader.cs
```diff
@@ -21,11 +21,11 @@ public WindowLoader(ContentManager contentManager, WndCallbackResolver wndCallba
             {
                 case SageGame.CncGenerals:
                 case SageGame.CncGeneralsZeroHour:
-                    contentManager.IniDataContext.LoadIniFile(@"Data\English\HeaderTemplate.ini");
+                    contentManager.IniDataContext.LoadIniFile($@"Data\{LanguageSetting.Current}\HeaderTemplate.ini");
```
paavohuhtala: Same here.

src/OpenSage.Game/LanguageSetting.cs
feliwir: Also the default language is "english" not "englisch". You've probably accidentally used the german version of "english"

feliwir:
Looking good to me aswell. Could merge, but want to wait for @tgjones feedback

src/OpenSage.Game/Utilities/LanguageUtility.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        /// <summary>
+        /// Detect the language for the current installation
+        /// </summary>
+        /// <param name="gameDefinition"></param>
+        /// <param name="rootDirectory"></param>
+        /// <returns>language as string</returns>
+        public static string ReadCurrentLanguage(IGameDefinition gameDefinition, string rootDirectory)
+        {
+            if (Environment.OSVersion.Platform == PlatformID.Win32Windows)
```
tgjones: Does this work? I think it should be Win32NT. Win32Windows was only for Windows95 and 98.

feliwir:
Thanks for the contribution!

PR #132, opened 2018-07-09
Resolve z fighting issues: Fixes #106

### LINKED ISSUE #106: Remedy z-fighting issues
Z-fighting with the terrain is prevalent with buildings with flat / low geometric features. Notable examples of buildings with these issues include GLA's command centre and USA's barracks and war factory.

image
The football stadium in `USA02.map` is especially bad; it has z-fighting at any realistic viewing distance and angle.

hashkitten:
The near clipping plane value was way too low, meaning that the depth buffer did not have enough precision to handle rendering everything correctly. Upping it to `4.0f` makes no discernible difference in gameplay; the only difference is if you zoom in very very close to a unit (to the point where you are almost on top of it), you may see the near polygons clipped. However you are never allowed to zoom in that far in practice. Here is what the `USA02.map` stadium looks like now:

zfightfix

### DIFF
--- a/src/OpenSage.Game/Graphics/Cameras/CameraComponent.cs
+++ b/src/OpenSage.Game/Graphics/Cameras/CameraComponent.cs
@@ -24,7 +24,7 @@ namespace OpenSage.Graphics.Cameras
             _game = game;
             _frustum = new BoundingFrustum(Matrix4x4.Identity);
 
-            NearPlaneDistance = 0.125f;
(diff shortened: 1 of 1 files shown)


tgjones:
This is great, thank you! That's been a long-time annoyance, but I never got round to looking into it.

I can imagine that some cinematic sequences make use of a (relatively) close near clipping plane, but we can always adjust this if we find anything incorrectly clipped.

PR #133, opened 2018-07-11
Support for mono DVI ADPCM compressed .wav files

hashkitten:
Some C&C Generals sounds are stored in a compressed format known as DVI (or IMA) ADPCM. One such example is `Data\Audio\Sounds\ctrustaa.wav`, which would previously cause a crash in the viewer when played. Since OpenAL does not support the format, we implement an on-the-fly decoder to unpack the compression into a linear PCM format that OpenAL can play. Note this commit only supports single channel files of this type, although supporting multiple channels should not be too hard (I haven't found any of these types of sounds yet though).

Uses
https://icculus.org/SDL_sound/downloads/external_documentation/wavecomp.htm
https://wiki.multimedia.cx/index.php/IMA_ADPCM
as references.

### DIFF
--- a/src/OpenSage.Game/Audio/AudioBuffer.cs
+++ b/src/OpenSage.Game/Audio/AudioBuffer.cs
@@ -22,43 +22,48 @@ namespace OpenSage.Audio
         public AudioBuffer(WavFile wavFile) : this()
         {
             int format = 0;
-
-            switch (wavFile.Channels)
+            if ((AudioFormatType) wavFile.AudioFormat != AudioFormatType.Microsoft)
             {
-                case 1:
-                    {
-                        switch (wavFile.BitsPerSample)
+                format = AL10.AL_FORMAT_MONO16;
+            }
+            else
+            {
+                switch (wavFile.Channels)
+                {
+                    case 1:
                         {
-                            case 8:
-                                format = AL10.AL_FORMAT_MONO8;
-                                break;
-                            case 16:
-                                format = AL10.AL_FORMAT_MONO16;
-                                break;
-                            default:
-                                throw new NotSupportedException("Invalid audio format!");
+                            switch (wavFile.BitsPerSample)
+                            {
+                                case 8:
+                                    format = AL10.AL_FORMAT_MONO8;
+                                    break;
+                                case 16:
+                                    format = AL10.AL_FORMAT_MONO16;
+                                    break;
+                                default:
+                                    throw new NotSupportedException("Invalid audio format!");
+                            }
                         }
-                    }
-                    break;
-                case 2:
-                    {
-                        switch (wavFile.BitsPerSample)
+                        break;
+                    case 2:
                         {
-                            case 8:
-                                format = AL10.AL_FORMAT_STEREO8;
-                                break;
-                            case 16:
-                                format = AL10.AL_FORMAT_STEREO16;
-                                break;
-                            default:
-                                throw new NotSupportedException("Invalid audio format!");
+                            switch (wavFile.BitsPerSample)
+                            {
+                                case 8:
+                                    format = AL10.AL_FORMAT_STEREO8;
+                                    break;
+                                case 16:
+                                    format = AL10.AL_FORMAT_STEREO16;
+                                    break;
+                                default:
+                                    throw new NotSupportedException("Invalid audio format!");
+                            }
                         }
-                    }
-                    break;
+                        break;
+                }
             }
 
-
-            AL10.alBufferData(_handle, format, wavFile.Buffer, wavFile.Size, wavFile.Fequency);
+            AL10.alBufferData(_handle, format, wavFile.Buffer, wavFile.Size, wavFile.Frequency);
             AudioSystem.alCheckError();
         }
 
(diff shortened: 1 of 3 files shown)


src/OpenSage.Game/Data/Wav/WavFile.cs
```diff
@@ -65,51 +135,118 @@ public sealed class WavFile
...
+        private int ByteToInt16(byte[] packed)
         {
-            _header = RiffHeader.Parse(reader);
+            // This is always little endian, unlike the C# builtin for unpacking a byte array.
+            return (packed[1] << 8) | packed[0];
+        }
 
-            if (new string(_header.ChunkId) != "RIFF" ||
-                new string(_header.Format) != "WAVE")
+        private byte[] ParseMicrosoft(BinaryReader reader, int size)
```
feliwir: It might be cleaner to implement a WavParser baseclass and add two derived classes of type PcmParser and AdpcmParser. I am pretty sure i saw other compressed .wav formats in BFME before, so it should be easier in the future to add new formats with that approach. But i can do that myself after merge if you don't want that refactoring now.

src/OpenSage.Game/Data/Wav/WavFile.cs
```diff
@@ -65,51 +135,118 @@ public sealed class WavFile
...
+        {
+            return reader.ReadBytes(size);
+        }
+        
+        private byte[] ParseDviAdpcm(BinaryReader reader, int size, int samplesPerBlock)
+        {
+            var imaIndexTable = new int[16] {
```
feliwir: this can only be of size 8 since the table is repeating. When accessing the table just a bitwise or with 0x7 (which will cause that 0x8 becomes 0 etc. See [here](https://github.com/dbry/adpcm-xq/blob/master/adpcm-lib.c#L150) for a reference of this technique.

src/OpenSage.Game/Data/Wav/WavFile.cs
```diff
@@ -65,51 +135,118 @@ public sealed class WavFile
...
+                    writer.Write((Int16) sample);
+                    for (int j = 0; j < (samplesPerBlock - 1) / 2; j++)
+                    {
+                        byte packed = reader.ReadByte();
+                        byte nibble_low = (byte) (packed & 0xF);
+                        byte nibble_high = (byte) (packed >> 4);
+                        foreach (var nibble in new byte[] { nibble_low , nibble_high })
+                        {
+                            stepTableIndex += imaIndexTable[nibble];
+                            if (stepTableIndex > 88) stepTableIndex = 88;
```
feliwir: Normally we don't use if Statements without brackets in OpenSage. But this is really just nitpicking, but @tgjones is really strict about this :smile:

src/OpenSage.Game/Data/Wav/WavFile.cs
```diff
@@ -65,51 +135,118 @@ public sealed class WavFile
...
             {
-                reader.BaseStream.Seek(_format.SubChunkSize - 16, SeekOrigin.Current);
+                _fact = WaveFact.Parse(reader);
             }
-                
-            _data = WaveData.Parse(reader);
 
-            if (new string(_data.SubChunkID) != "data")
+            _data = WaveData.Parse(reader);
+            switch ((AudioFormatType) _format.AudioFormat)
```
feliwir: As i said this would not be necessary if you created a baseclass for the decoders and abstracted this part away.

src/OpenSage.Game/Data/Wav/WavFile.cs
```diff
@@ -65,51 +135,118 @@ public sealed class WavFile
...
+                    int stepTableIndex = reader.ReadByte();
+                    byte reserved = reader.ReadByte(); // unused, commonly 0
+                    int step = imaStepTable[stepTableIndex];
+                    writer.Write((Int16) sample);
+                    for (int j = 0; j < (samplesPerBlock - 1) / 2; j++)
+                    {
+                        byte packed = reader.ReadByte();
+                        byte nibble_low = (byte) (packed & 0xF);
+                        byte nibble_high = (byte) (packed >> 4);
+                        foreach (var nibble in new byte[] { nibble_low , nibble_high })
```
feliwir: Also this part is quite inefficient since it always creates a List on the Heap. As @paavohuhtala pointed out to me C# added a stackalloc feature that allocates memory on the stack, which works like this:
```csharp
Span<byte> nibbles = stackalloc byte[2];
nibbles[0] = nibble_low;
nibbles[1] = nibble_high;
```

This should be much faster and much lower overhead. See [here](http://adamsitnik.com/Span/)  for more information about Span<T> with stackalloc feature. I tested this code locally and it does work

src/OpenSage.Game/Audio/AudioBuffer.cs
```diff
@@ -22,43 +22,48 @@ public AudioBuffer()
         public AudioBuffer(WavFile wavFile) : this()
         {
             int format = 0;
-
-            switch (wavFile.Channels)
+            if ((AudioFormatType) wavFile.AudioFormat != AudioFormatType.Microsoft)
```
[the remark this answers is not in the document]
hashkitten: Unfortunately BitsPerSample represents the compressed bits per sample, not the uncompressed bits per sample, so for ADPCM it is 4. But you're right, it's probably not the right place to put this code, so I have opted to make BitsPerSample a method that returns the 'true' uncompressed bits per sample rather than polluting AudioBuffer.

hashkitten:
Thanks for the review. Hopefully I have addressed these issues with the latest commit (`3b8be5a`).

PR #136, opened 2018-07-27
Add ability to change the house color from GameObjects

MarkusHiller:
coloring

We are now abel to change the HouseColors bay setting Player.Color.

### DIFF
--- a/src/OpenSage.Game/Content/ContentManager.cs
+++ b/src/OpenSage.Game/Content/ContentManager.cs
@@ -239,7 +239,7 @@ namespace OpenSage.Content
             var objectDefinition = IniDataContext.Objects.FirstOrDefault(x => x.Name == typeName);
             if (objectDefinition != null)
             {
-                return new GameObject(objectDefinition, this);
+                return new GameObject(objectDefinition, this, _game.CivilianPlayer);
             }
             else
             {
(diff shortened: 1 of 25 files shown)


feliwir:
The screenshot is looking great, thanks for this change. As a general design idea (for the future maybe): It might be better to apply the house color in the shader (e.g. FixedFunction.cs) instead of changing the VertexMaterial. So the MeshConstants struct could potentially look like this: 
```csharp
        public struct MeshConstants
        {
            public /* bool */ uint SkinningEnabled;
            public uint NumBones;
            public /* bool */ uint HasHouseColor;
            public Vector3 HouseColor;
        }
```
The HouseColor could then be used in the Fragment/Pixelshader when HasHouseColor is set.
The benefit of this approach would be that models with different house colors could still be rendered with a single draw call (through [instancing](https://learnopengl.com/Advanced-OpenGL/Instancing)). Since we aren't using instancing at the moment i don't see any issue with doing it like you do, but it's something we need to keep in mind for the future. Waiting for opinions of the others on this before merging.

feliwir:
So i did the changes i've described above. Also now the color that was set inside the skirmish menu is used. Since i wrote code of this myself now it would be nice if some could have a short look

tgjones:
Replaced by #237. @MarkusHiller thank you for your work on this - I just extended it a bit to apply the house colour in the shader instead of at load time.

PR #137, opened 2018-07-28
Added GUI sounds

feliwir:
Plays the correct sound when pressing a Button in Generals. Loads the corresponding settings from the .ini files. Split WavParsers into seperate files.

### DIFF
--- a/src/OpenSage.Game/Data/Wav/DviApcmParser.cs
+++ b/src/OpenSage.Game/Data/Wav/DviApcmParser.cs
@@ -6,19 +6,11 @@ namespace OpenSage.Data.Wav
 {
     internal class DviAdpcmParser : WavParser
     {
-        private int ByteToInt16(byte[] packed)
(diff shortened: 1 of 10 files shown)


src/OpenSage.Game/Data/Wav/DviApcmParser.cs
```diff
@@ -0,0 +1,92 @@
+﻿using System;
+using System.IO;
+using OpenSage.Mathematics;
+
+namespace OpenSage.Data.Wav
+{
+    internal class DviAdpcmParser : WavParser
+    {
+        private int ByteToInt16(byte[] packed)
```
paavohuhtala: If it actually takes **two bytes** and returns an **int32**, then `ByteToInt16` isn't really accurate. 🤔 Maybe call it something like `Int16FromLittleEndianBytes` and actually cast it to int16?

PR #138, opened 2018-07-29
Gui improvements

feliwir:
- Fix hovered listbox item being displayed correctly
- Fixed text alignment of listbox items being too much on the left (i think this is hardcoded. I did choose 10px for now)
- Fixed that comboxes aren't closing when pressing anywhere on the screen
- Fixed that expand won't expand when clicking the textarea
- Added closing functionality when exiting the game through the interface. (the viewer does not close when exiting for obvious reasons)

Screenshot of the skirmish menu:
grafik

### DIFF
--- a/src/OpenSage.Game/Gui/Wnd/Controls/Control.cs
+++ b/src/OpenSage.Game/Gui/Wnd/Controls/Control.cs
@@ -74,9 +74,13 @@ namespace OpenSage.Gui.Wnd.Controls
             }
         }
 
-        // TODO: Exclude BorderWidth.
-        public Rectangle ClientRectangle => new Rectangle(0, 0, _bounds.Width, _bounds.Height);
-        public Size ClientSize => new Size(_bounds.Width, _bounds.Height);
+        public Rectangle ClientRectangle => new Rectangle(BorderWidth,
+                                                          BorderWidth,
+                                                          _bounds.Width- BorderWidth,
+                                                          _bounds.Height- BorderWidth);
+
+        public Size ClientSize => new Size(_bounds.Width - BorderWidth,
+                                           _bounds.Height - BorderWidth);
 
         public int Left
         {
@@ -440,6 +444,7 @@ namespace OpenSage.Gui.Wnd.Controls
                     break;
 
                 case WndWindowMessageType.MouseUp:
+                    CloseOpenComboboxes(this);
                     IsMouseDown = false;
                     break;
             }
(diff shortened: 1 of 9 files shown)


src/OpenSage.Game/Gui/Wnd/Controls/ListBox.cs
```diff
@@ -289,17 +300,36 @@ public int MaxDisplay
             }
         }
 
-        public void UpdateSelectedItem()
+        private void SetSelectedItem()
```
paavohuhtala: Since this is not a setter method (and doesn't take a parameter) I think the previous name was better.

When I see a call to `SetSelectedItem()` without an argument, the first question that comes to my mind is "Set it to what?"

src/OpenSage.Game/Gui/Wnd/Controls/Control.cs
```diff
@@ -480,6 +485,17 @@ public static IEnumerable<Control> GetSelfAndDescendants(Control baseControl, bo
                 }
             }
         }
+
+        //TODO: check if this can be made more efficient
+        private static void CloseOpenComboboxes(Control sender)
```
paavohuhtala: We _could_ have a per-window cache of open comboboxes, but I don't see this being a bottleneck any time soon.

src/OpenSage.Game/Gui/Wnd/Controls/ListBox.cs
```diff
@@ -289,17 +300,36 @@ public int MaxDisplay
...
+                else
+                {
+                    item.BackgroundImage = null;
+                }
+            }
+        }
+
+        private void SetHoveredItem()
+        {
+            for (var i = 0; i < _items.Length; i++)
```
paavohuhtala: If we want to get fancy and micro-optimize everything, I think this (and `SetSelectedItem`) could be implemented without a loop. Instead of using `_hoveredIndex` pass the previous and new hovered index as arguments, and then just do something like this:

```csharp

if (previousIndex == nextIndex) return;

var previousItem = Controls[previousIndex] as ListBoxItem;
previousItem.BackgroundImage = previousIndex === _selectedIndex ? SelectedItemBackgroundImage : null;

var nextItem = Controls[nextIndex] as ListBoxItem;
nextItem.BackgroundImage = nextIndex === _selectedIndex ? SelectedItemHoverBackgroundImage : null;
```

(Untested, feel free to use full `if`s if you think it makes it more readable.)

PR #142, opened 2018-10-27
Handle escape press in credits

### LINKED ISSUE #141: Credits Issue
When playing through the credits, I noticed that there is no way to exit back to the main menu.

feliwir:
Fixes #141


src/OpenSage.Game/Graphics/Effects/EffectParameter.cs
```diff
@@ -55,7 +55,10 @@ public void ApplyChanges(CommandList commandEncoder)
                 return;
             }
 
-            commandEncoder.SetGraphicsResourceSet(_slot, _data);
+            if (_data!=null)
```
paavohuhtala:
```suggestion
            if (_data != null)
```

PR #143, opened 2018-10-27
Replace commandline parser

feliwir:
The commandline parser we used previously was declared as "System.Commandline" parser, but was actually a private project (see [here](https://www.nuget.org/packages/System.CommandLine/1.0.0.4)). The project is still available [here](https://github.com/deniszykov/commandline).

Instead of changing to the new repository i decided to switch to the much more used commandline parser. It has a much nicer api and allows short/long options etc. It can directly parse Enums from the arguments. Overall i think it's the better choice for future development.

### DIFF
--- a/src/OpenSage.Launcher/OpenSage.Launcher.csproj
+++ b/src/OpenSage.Launcher/OpenSage.Launcher.csproj
@@ -14,8 +14,8 @@
   </ItemGroup>
   
   <ItemGroup>
+    <PackageReference Include="CommandLineParser" Version="2.3.0" />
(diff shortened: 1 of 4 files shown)


tgjones:
Just to clear one thing up: the System.CommandLine package we have been using is this one: https://dotnet.myget.org/feed/dotnet-corefxlab/package/nuget/System.CommandLine

It was an official (prototype) project from Microsoft, but it’s been cancelled now.

So i’m happy to switch to whatever is the current best choice.

feliwir:
Alright, then i’ll merge this

PR #150, opened 2018-11-05
[WIP] Compile launcher and viewer into native binaries

feliwir:
Use CoreRT to compile the Viewer and the Launcher into a single native executable for publishing. This reduces the overall size of the binaries and should allow greater speed. There are still some open issues though:

- [ ] Evaluate if the `rd.xml` (used for reflection) can be put into a shared project (OpenSage.Game).
- [ ] Try building OpenSage.Game as a shared native library and reuse that for the Viewer and Launcher
- [ ] Measure if there are any performance benefits
- [ ] Try to avoid reflection as much as possible
- [ ] Fix shader loading
- [ ] Fix SharpDx reflection issue

This PR compiles the project so far, but gives a runtime error when we are trying to load the embedded shaders:
```
Unhandled Exception: System.ArgumentNullException: Value cannot be null.
Parameter name: key
   at OpenSage.Launcher!<BaseAddress>+0x290a79
   at OpenSage.Launcher!<BaseAddress>+0x8902f7
   at OpenSage.Graphics.Effects.Effect..ctor(GraphicsDevice, String, VertexLayoutDescription[]) + 0x25d
   at OpenSage.Graphics.Effects.EffectLibrary..ctor(GraphicsDevice) + 0x7b
   at OpenSage.Content.ContentManager..ctor(Game, FileSystem, GraphicsDevice, SageGame, WndCallbackResolver) + 0x80e
   at OpenSage.Game..ctor(IGameDefinition, FileSystem, GamePanel) + 0x222
   at OpenSage.Launcher.Program.Run(Options) + 0x1da
   at CommandLine.ParserResultExtensions.WithParsed[T](ParserResult`1, Action`1) + 0x33
   at OpenSage.Launcher.Program.Main(String[]) + 0x7e
   at OpenSage.Launcher!<BaseAddress>+0x9455de
```

### DIFF
--- a/src/OpenSage.Game/Data/Ini/Parser/IniParser.Attributes.cs
+++ b/src/OpenSage.Game/Data/Ini/Parser/IniParser.Attributes.cs
@@ -58,7 +58,7 @@ namespace OpenSage.Data.Ini.Parser
         }
 
         public T ParseAttributeEnum<T>(string label)
-            where T : struct
+            where T : struct, IConvertible
         {
             return ParseAttribute(label, x => ParseEnum<T>(x));
         }
--- a/src/OpenSage.Game/Data/Ini/Parser/IniParser.Attributes.cs
+++ b/src/OpenSage.Game/Data/Ini/Parser/IniParser.Attributes.cs
@@ -58,7 +58,7 @@ namespace OpenSage.Data.Ini.Parser
         }
 
         public T ParseAttributeEnum<T>(string label)
-            where T : struct, IConvertible
+            where T : Enum
         {
             return ParseAttribute(label, x => ParseEnum<T>(x));
         }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,28 @@
+{
+   // Use IntelliSense to find out which attributes exist for C# debugging
+   // Use hover for the description of the existing attributes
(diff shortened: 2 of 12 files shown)


src/OpenSage.Game/Data/Ini/Parser/IniParser.Attributes.cs
```diff
@@ -37,7 +37,7 @@ public string ParseAttributeIdentifier(string label)
         }
 
         public T ParseAttributeEnum<T>(string label)
-            where T : struct
+            where T : struct, IConvertible
```
tgjones: Isn’t there an Enum generic constraint now?

feliwir: Oh that would be awesome. Didn't know that this existed. Will change.

feliwir: Can't use that Enum constraint, because it doesn't work with the BitArray<T> stuff. Enum is a nullable type. Combining "Enum, struct" is not sufficient either.

tgjones: Would it work if you changed the constraints on BitArray as well?

tgjones:
Closing in favour of #272.

PR #158, opened 2018-11-13
Parse shader material ids, skip corrupted and unreferenced file

Tarcontar:
Parse shader material ids. Skip file "readonly-0-rev-2-lwbanhwitchk.w3d" in unit tests, since it is corrupted and not referenced by any ini file. 
image

### DIFF
--- a/src/OpenSage.Game.Tests/Data/W3d/W3dFileTests.cs
+++ b/src/OpenSage.Game.Tests/Data/W3d/W3dFileTests.cs
@@ -42,6 +42,7 @@ namespace OpenSage.Tests.Data.W3d
                     case "lwbanhnazgul.w3d":
                     case "lwbanhwtchkng.w3d":
                     case "psupplies04.w3d":
+                    case "readonly-0-rev-2-lwbanhwitchk.w3d":
                     case "wbcave_d2a.w3d":
                     case "wbcave_d2c.w3d":
                         return; // Corrupt, or unreferenced and contain chunks that don't exist elsewhere.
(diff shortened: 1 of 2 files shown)


tgjones:
I need to look at what to do when there is more than one shader material per material pass. Currently the rendering code assumes there is only one. How many models have this?

Tarcontar:
I only found one model with that, and even that one might not be used ingame.

tgjones:
I’m bidding for a RotWK DVD on eBay, it’s about time I can look at these issues myself...

tgjones:
And now my RotWK DVD is (hopefully) on its way to me, then I'll look at this.

tgjones:
There are only two models where `W3D_CHUNK_SHADER_MATERIAL_ID` contains more than one shader material ID:

* kbpostgaten_al.w3d
* kbpostgaten_am.w3d

It looks like there's one shader material ID for each triangle. We *could* support that, but... neither of these models are referenced by any ini files, so let's not.

Could you add these two models to the test exclusion list, and then remove the other changes to support multiple shader material IDs?

Tarcontar:
Yes I could do that. But what exactly do u mean by "other changes to support multiple shader material IDs"? I think the changes in the test file should remain the same and also the parsing should remain as it is.

tgjones:
I mean, since we'll ignore the only two files that have multiple shader material IDs, we don't need to change the parsing code to support multiple shader material IDs. Because if we're really going to support that, we also need to change the rendering code, which is a lot more work.

Tarcontar:
Well the current changes have no impact to any other part of the engine. Just that the chunk is read properly and no exception is thrown. In my opinion this is the better approach.

tgjones:
In general, when the engine doesn't support a particular thing in w3d, I throw an exception during parsing. It's the same here - the engine doesn't support multiple shader material IDs, and no models actually used by any game need that feature, so I think it's safer to keep it as it was.

Tarcontar:
Ok, ill change it

Tarcontar:
done, should be ready to merge now

tgjones:
Thank you!

PR #167, opened 2018-11-18
Viewer improvements

paavohuhtala:
* Don't draw the game view when there's not enough space. Fixes a crash when the viewer window is really small.
* Implement a text view for INI files.
  * This currently uses a normal unformatted text control. This means the text cannot be highlighted or copied. We'll migrate to a readonly multiline input when ImGui.NET has solved [this issue](https://github.com/mellinoe/ImGui.NET/issues/82).

image

* Show a notification when a game view has input focus

image

* Add a context menu option to copy INI contents to clipboard.
* Default to text view when the file doesn't contain any known objects (and hide the view selector).

### DIFF
--- a/src/OpenSage.Viewer/UI/MainForm.cs
+++ b/src/OpenSage.Viewer/UI/MainForm.cs
@@ -174,6 +174,13 @@ namespace OpenSage.Viewer.UI
 
                 ImGui.Text(_contentView.DisplayName);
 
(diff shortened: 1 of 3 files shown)


feliwir:
Looks good to me. I've tested it locally and everything works as expected.

tgjones:
Good stuff, thank you!

PR #169, opened 2018-11-18
FPS massaging

Qibbi:
Adds an fps counter to game view.
Adds a vsync option. (vsync is on by default from prev version)
Changed some math lib methods to static to prevent defensive copies to be created.

I wanted to split these but I have no idea how to ;D


src/OpenSage.Viewer/UI/Views/GameView.cs
```diff
@@ -56,6 +56,10 @@ public override void Draw(ref bool isGameViewFocused)
             {
                 isGameViewFocused = true;
             }
+            ImGui.SetCursorScreenPos(windowPos + new Vector2(1.0f));
+            ImGui.TextColored(Vector4.UnitW, $"{ImGui.GetIO().Framerate:N2} FPS");
```
tgjones: I didn’t know you can do formatting like this in an interpolated string. Very cool.

src/OpenSage.Game/Graphics/Cameras/RtsCameraController.cs
```diff
@@ -185,7 +185,7 @@ void ICameraController.UpdateCamera(CameraComponent camera, in CameraInputState
                 new Vector3(0, 0, cameraHeight),
                 new Vector3(0, 1, cameraHeight),
                 new Vector3(1, 0, cameraHeight));
-            var toCameraIntersectionDistance = toCameraRay.Intersects(ref plane).Value;
```
tgjones: This is great - I hadn’t appreciated the extra cost of “in” when calling instance methods on those parameters.

I wonder, instead of changing everything to static methods, should we change our maths struct types to “readonly struct”? I prefer immutable maths types anyway.

Qibbi: That would have the same effect, yes. Though sometimes setting something on the fly can be nice instead of creating a new one (if you calculate a vector for something but you want the Z value to be 1.0f for example). One would have to look if there are usecases for that though.

src/OpenSage.Viewer/UI/Views/GameView.cs
Qibbi: It's neat, though I should have really cached the string instead of doing that twice for the dropshadow.

src/OpenSage.Game/Graphics/Cameras/RtsCameraController.cs
tgjones: I don't think we often mutate struct fields. If we did - for example, `Ray.Position`, we could always have a `WithPosition(in Vector3 newPosition)` instance method that returns a new `Ray`. Would you mind making this change? I think we should make these `struct`s readonly:

* `BoundingBox`
* `BoundingSphere`
* `ColorRgb`
* `ColorRgba`
* `ColorRgbaF`
* `ColorRgbF`
* `Line2D`
* `ColorRgbF`
* `Matrix2x2`
* `Matrix4x3`
* `Point2D`
* `ColorRgbF`
* `Ray`
* `Rectangle`
* `RectangleF`
* `Size`
* `SizeF`
* `Triangle`

(And then we can go back to using instance methods on these types.)

PR #171, opened 2018-11-19
Fix tile blending in BFME

tgjones:
I figured what the previous `Unknown` bytes were for. The `Blends`, `ThreeWayBlends`, and `CliffTextures` data seems to be 32-bit between versions 14 and 24. And then after version 24, oddly, went back to 16-bit.


tgjones:
Tile blending still isn’t right for BFME, but this is an improvement on what we have now.

PR #109, opened 2018-04-21
[WIP] Threaded loading

paavohuhtala:

PR #172, opened 2018-11-19
Subsystems

paavohuhtala:
* Implement rudimentary subsystem support. Subsystems are a concept in BFME and newer games, which are used to load related configuration files.
* Tweak `LoadSubsystem` parsing to be order-preserving

This PR doesn't integrate subsystems properly into the engine yet. Most subsystems are immediately loaded in `ContentManager` and most manual INI loading calls still remain. Not every subsystem has been added yet, and I've only looked at BFME2 and C&C3 so far.


paavohuhtala:
Supersedes #109, mostly.

PR #173, opened 2018-11-23
Implement parsing of W3D model data in C&C3 asset streams

tgjones:


src/OpenSage.Game/Graphics/ModelMesh.cs
```diff
@@ -71,9 +72,7 @@ public sealed class ModelMesh : DisposableBase
 
             _effect = effect;
 
-            _vertexBuffer = AddDisposable(graphicsDevice.CreateStaticBuffer(
-                vertices,
-                BufferUsage.VertexBuffer));
+            _vertexBuffer = AddDisposable(graphicsDevice.CreateVertexBuffer(vertexData));
```
feliwir: Indices and vertices should use the same `CreateXXX` function

tgjones: But `CreateStaticBuffer` takes a `T[]`, while `CreateVertexBuffer` takes a `byte[]`. I could name them both `CreateStaticBuffer`, if you'd prefer?

BSG-75:
I tried it with RA3 and it crashed...
because it tried to load "`Data\\INI\\Default\\subsystemlegend.ini"`, and `_filesystem.GetFile()` returned `null`

With my C&C:3 it crashes as well with other kind of ini errors...

tgjones:
Yes, it doesn’t work in the viewer yet (and rendering isn’t yet implemented anyway). I’ve just been running the ManifestFileTests while working on parsing.

Qibbi:
For C&C3 and onward if you want support versions which aren't just the latest ones it would be important to identify an asset type not just by its type id but also it's type hash.

Qibbi:
Also shouldn't OpenSage.Data.StreamFS.AssetReaders.AssetReader be an interface? It only has an abstract read only property and an abstract method.

tgjones:
@qibbi interesting - does the type hash change per version? I didn’t notice that. I’m supporting multiple versions now, but only by checking the host game context, which is a bit hacky.

tgjones:
@BSG-75 Oops - turns out the C&C3 and RA3 ini errors were because of something @paavohuhtala is working on. I've done a temporary workaround, so you can switch to C&C3 and RA3 in the viewer.

You still can't view w3d assets for those games, though, I still need to work more on that.

Qibbi:
The change from TW to KW in W3DMesh is that they added a "GeometryPickable" bool. It doesn't change the assets size (as before it had 3 1-byte members, now it has 4, but with padding it is the same size in the end).

Issue #174, opened 2018-11-24
Viewer currently only searches for system variables on detecting installations


Tarcontar:
At the moment the viewer and specifically the method to locate game installations only searches in the system environment variables for installation paths, would be neat if it could also search in the user environment variables.
This could be done by setting the following parameter in the function:

String path = Environment.GetEnvironmentVariable("BFME_PATH", EnvironmentVariableTarget.User);

In InstallationLocator.cs.

tgjones:
Fixed by #175.

PR #175, opened 2018-11-24
Fix for issue #174: installation locator

Tarcontar:
I only use windows, but dont do always a clean new installation of the games on my various systems, I simply copy them over. So it would be neat to simply set the installation path as on linux.
Also added a check if the found path exists, so the viewer does not crash if the installation folder was moved or deleted.

### DIFF
--- a/src/OpenSage.Game/Data/InstallationLocator.cs
+++ b/src/OpenSage.Game/Data/InstallationLocator.cs
@@ -67,10 +67,17 @@ namespace OpenSage.Data
     {
         public IEnumerable<GameInstallation> FindInstallations(IGameDefinition game)
         {
-            String path = Environment.GetEnvironmentVariable(game.Identifier.ToUpperInvariant() + "_PATH");
+            var identifier = game.Identifier.ToUpperInvariant() + "_PATH";
+            var path = Environment.GetEnvironmentVariable(identifier);
 
-            if (path == null)            
+            if (path == null)
+            {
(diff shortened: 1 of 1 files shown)


src/OpenSage.Game/Data/InstallationLocator.cs
```diff
@@ -67,10 +67,17 @@ public class EnvironmentInstallationLocator : IInstallationLocator
...
         public IEnumerable<GameInstallation> FindInstallations(IGameDefinition game)
         {
-            String path = Environment.GetEnvironmentVariable(game.Identifier.ToUpperInvariant() + "_PATH");
+            var identifier = game.Identifier.ToUpperInvariant() + "_PATH";
+            var path = Environment.GetEnvironmentVariable(identifier);
 
-            if (path == null)            
+            if (path == null)
+            {
+                path = Environment.GetEnvironmentVariable(identifier, EnvironmentVariableTarget.User);
```
tgjones: Interesting - does this find values that the method above doesn't? I thought on Windows the current process inherited the user's environment variables?

tgjones: (because `Environment.GetEnvironmentVariable(string)` is the same as calling `Environment.GetEnvironmentVariable(string, EnvironmentVariableTarget.Process)`.)

Tarcontar: Yea, the user path variables are not found by the current implementation.

tgjones: Okay, looks like a good change then.

PR #176, opened 2018-11-25
Further Bfme ini parsing

Tarcontar:


src/OpenSage.Game/Logic/Object/Draw/W3dHordeModelDraw.cs
```diff
@@ -9,6 +10,39 @@ public class W3dHordeModelDrawModuleData : W3dModelDrawModuleData
...
+            {
+                { "StaticModelLODMode", (parser, x) => x.StaticModelLODMode = parser.ParseBoolean() },
+                { "LodOptions", (parser, x) => x.LODOptions.Add(LODOption.Parse(parser)) }
+            });
+
+        public bool StaticModelLODMode { get; internal set; }
+        public List<LODOption> LODOptions { get; internal set; } = new List<LODOption>();
+    }
+
+    public sealed class LODOption
```
tgjones: Could you rename this to `LodOption`, to match .NET naming convention that acronyms with three letters or more should be Pascal-cased?

src/OpenSage.Game/Logic/Object/Draw/W3dHordeModelDraw.cs
```diff
@@ -9,6 +10,39 @@ public class W3dHordeModelDrawModuleData : W3dModelDrawModuleData
...
         internal static new readonly IniParseTable<W3dHordeModelDrawModuleData> FieldParseTable = W3dModelDrawModuleData.FieldParseTable
-            .Concat(new IniParseTable<W3dHordeModelDrawModuleData>());
+            .Concat(new IniParseTable<W3dHordeModelDrawModuleData>
+            {
+                { "StaticModelLODMode", (parser, x) => x.StaticModelLODMode = parser.ParseBoolean() },
+                { "LodOptions", (parser, x) => x.LODOptions.Add(LODOption.Parse(parser)) }
+            });
+
+        public bool StaticModelLODMode { get; internal set; }
+        public List<LODOption> LODOptions { get; internal set; } = new List<LODOption>();
```
tgjones: Please rename this to `LodOptions`.

src/OpenSage.Game/Logic/Object/Draw/W3dHordeModelDraw.cs
```diff
@@ -9,6 +10,39 @@ public class W3dHordeModelDrawModuleData : W3dModelDrawModuleData
...
+
+        internal static readonly IniParseTable<LODOption> FieldParseTable = new IniParseTable<LODOption>
+            {
+                { "AllowMultipleModels", (parser, x) => x.AllowMultipleModels = parser.ParseBoolean() },
+                { "MaxRandomTextures", (parser, x) => x.MaxRandomTextures = parser.ParseInteger() },
+                { "MaxRandomAnimations", (parser, x) => x.MaxRandomAnimations = parser.ParseInteger() },
+                { "MaxAnimFrameDelta", (parser, x) => x.MaxAnimFrameDelta = parser.ParseInteger() }
+            };
+
+        public ModelLevelOfDetail LOD { get; internal set; }
```
tgjones: Please rename this to `Lod`.

src/OpenSage.Game/Logic/Object/Draw/W3dHordeModelDraw.cs
```diff
@@ -9,6 +10,39 @@ public class W3dHordeModelDrawModuleData : W3dModelDrawModuleData
...
+
+        public bool AllowMultipleModels { get; internal set; }
```
tgjones: Do these need to be `internal set`? Could they be `private set`? I've noticed that in a few other places too.

src/OpenSage.Game/Logic/Object/SpecialPower/SiegeDepolySpecialPower.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections.Generic;
```
tgjones: This file's filename has a typo.

src/OpenSage.Game/Logic/Object/Draw/W3dHordeModelDraw.cs
Tarcontar: Saw that in a few places, and used internal by default for new objects. Should I use private instead?

tgjones: I thought I only used `internal` when it actually needed to be accessed by something else, like a subclass or another class. Yes, `private` is better if it doesn't need to be accessed by anything else.

Tarcontar: ok, so should i change that for all the objects, also for the ones from the old PR?

tgjones: If you have the energy to do so, yes please.

Tarcontar:
done, anything else?

tgjones:
Thank you!

PR #178, opened 2018-11-26
Render shadows

tgjones:


tgjones:
Merging this to avoid divergence with other PRs. Shadows are working quite well, but there is still a problem with the shadow frustum calculation meaning that shadows are lower quality than they should be. So, shadows are disabled by default and can be enabled in the map viewer, until I do another PR that fixes frustum calculation.

PR #182, opened 2018-11-28
Viewer file search improvements

paavohuhtala:
* Cache search results. They're now only computed when the search changes. This has a fairly major (positive) FPS impact.
* Fix null-terminator issues with input boxes (thanks @Qibbi!)
* Clear search box when installation changes.
* Small optimizations

image


feliwir:
Looks good to me

PR #189, opened 2018-12-02
Unit selection 1.5

paavohuhtala:
Based on #183 

* Implement selection barks (_Are there not better tools?_)
* Modify `SelectionSystem` so that unit selection events are passed through `NetworkMessageBuffer` and `OrderProcessor`. 
* Fix Generals replay menu crash
* Parse `ScreenRectangle` order arguments and `DrawBoxSelection` orders from replays
* Process local orders locally


tgjones:
Looks good to me, thanks!

PR #191, opened 2018-12-03
Fix transparency bug in game views in Viewer

tgjones:
... by rendering to an intermediate target, and ignoring alpha when blitting from that to the output framebuffer.


feliwir:
Looks good to me

PR #194, opened 2018-12-03
fix missing audio stuff that prevents bfme2 from loading

Tarcontar:


src/OpenSage.Game/Audio/AudioSystem.cs
```diff
@@ -39,7 +39,7 @@ public AudioSystem(Game game) : base(game)
                 case SageGame.CncGeneralsZeroHour:
                 case SageGame.Bfme:
                 //TODO: complete audio settings for BFME
```
tgjones: Can this TODO be deleted now?

Issue #31, opened 2018-01-02
Support loading all campaign maps


paavohuhtala:
Tracking issue for which campaign maps load and which don't. Loadability just means that the map doesn't crash the engine.

# Generals

## China
- [x] `CHI01`
- [x] `CHI02`
- [x] `CHI03`
  - Crashes on load, because `TerrainPatchIndexBufferCache` tries to create an index buffer with the length of 0. The patch has the size of 1x17 and `CalculateNumIndices(1, 17)` returns 0.
- [x] `CHI04`
- [x] `CHI05`
- [x] `CHI06`
- [x] `CHI07`
- [x] `CHI08` (a cut, unfinished map)
- [x] `CHIfinal`
## GLA
- [x] `GLA01`
- [x] `GLA02`
- [x] `GLA04`
- [x] `GLA05`
- [x] `GLA06`
- [x] `GLA07`
- [x] `GLA08`

## USA
- [x] `USA01`
- [x] `USA02`
- [x] `USA03`
- [x] `USA04`
- [x] `USA05`
- [x] `USA06`
- [x] `USA07`
- [x] `USA08`

## Misc
- [x] `Training01`
- [x] `ShellMap1` 
- [x] `ShellMap1Low`

# Zero Hour

## Debug maps
- [x] `AllBuildingsAllSidesUnitTest_Save`
  - Uses the nonexistent model "Art\W3D\NONE.W3D", causing a crash.
- [x] `_Art Review New Units`
- [x] `BUG_SavedGameandEnabledFolders`

smx-smx:
bumping the TerrainPatchSize width and height by 1 makes CHI03 load but it needs a proper fix.
The source of those values is MapLoader.CreatePatches, and in particular the patchBounds variable

Continous:
"NONE.W3D" may be their hacky way of saying null?

Qibbi:
That is correct. It's also used in quite a few objects when no model should be drawn in a specific state.

feliwir:
Fixed with #199

PR #199, opened 2018-12-04
Allow null model for animation

tgjones:
Final fix for #31.

Incidentally, `AllBuildingsAllSidesUnitTest_Save.map` is a great stress test, which OpenSAGE currently... fails :)

2018-12-03 all buildings map


feliwir:
Looks good to me

PR #206, opened 2018-12-05
Fix GLSL/C# mismatches

feliwir:
- Adds a Bool4 type in C# that matches the 4 byte bools in GLSL.
- Change ConstantBuffer to use the unsafe sizeof operator which returns the "real" size of the underlying type (only relevant for `bool` type)


tgjones:
Thanks for this!

PR #210, opened 2018-12-07
Some missing ini stuff for bfme to pass unit test

Tarcontar:


src/OpenSage.Game/Data/Ini/IniDataContext.cs
```diff
@@ -78,8 +93,18 @@ public sealed class IniDataContext
...
+        public Pathfinder Pathfinder { get; set; }
+
+        [AddedIn(SageGame.Bfme)]
+        public List<PlayerAIType> PlayerAITypes { get; } = new List<PlayerAIType>();
+
         public List<PlayerTemplate> PlayerTemplates { get; } = new List<PlayerTemplate>();
         public List<Rank> Ranks { get; } = new List<Rank>();
+
+        [AddedIn(SageGame.Bfme)]
+        public RegionCampain RegionCampain{ get; set; }
```
tgjones: Spotted a typo here, should be `RegionCampaign`.

Tarcontar: thats how its spelled inside the ini files, thought it would be easier to find in the code this way, but could rename it

src/OpenSage.Game.Tests/Data/Ini/IniFileTests.cs
```diff
@@ -19,37 +19,45 @@ public IniFileTests(ITestOutputHelper output)
...
             {
                 GameDefinition.FromGame(SageGame.CncGenerals),
-                GameDefinition.FromGame(SageGame.CncGeneralsZeroHour)
+                GameDefinition.FromGame(SageGame.CncGeneralsZeroHour),
+                GameDefinition.FromGame(SageGame.Bfme),
+                //GameDefinition.FromGame(SageGame.Bfme2),
             };
 
-            InstalledFilesTestData.ReadFiles(".ini", _output, gameDefinitions, entry =>
+            foreach(var gameDefinition in gameDefinitions)
```
tgjones: Why this change?

src/OpenSage.Game/Data/Ini/WindowTransition.cs
```diff
@@ -87,4 +124,62 @@ public enum WindowTransitionStyle
         [IniEnum("CONTROLBARARROW")]
         ControlBarArrow
     }
+
+    [AddedIn(SageGame.Bfme)]
+    public enum TransitionType
```
tgjones: Are you sure these are a different thing from WindowTransitionStyle?

src/OpenSage.Game.Tests/Data/Ini/IniFileTests.cs
Tarcontar: this is to get the right game definition here:
var dataContext = new IniDataContext(entry.FileSystem, gameDefinition.Game);
the old approach was to use always SageGame.Generals, but since things like animations are game dependant this wouldnt work, so we need the proper game definition here.

src/OpenSage.Game/Data/Ini/WindowTransition.cs
Tarcontar: True, this seems to be the same stuff, 
will fix it

Tarcontar:
Combined the two enums

PR #217, opened 2018-12-11
Prevent initialization of AudioSystem for ra3/uprising/cnc4.

BSG-75:
Prevent initialization of `AudioSystem` for games which we haven't implemented their audio settings yet, so OpenSage Viewer won't crash when switching to games like RA3 or Uprising.

### DIFF
--- a/src/OpenSage.Game/Audio/AudioSystem.cs
+++ b/src/OpenSage.Game/Audio/AudioSystem.cs
@@ -33,7 +33,8 @@ namespace OpenSage.Audio
                 case SageGame.Ra3Uprising:
                 case SageGame.Cnc4:
                     // TODO
-                    break;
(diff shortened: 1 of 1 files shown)


paavohuhtala:
Thanks, this is good stuff!

I think you should remove the logging call, and just replace it with a TODO comment like the others in the same file. Not because it's unwanted, but because we're integrating a proper logging solution soon (hopefully by the end of the month!), and `Debug.WriteLine` calls are somewhat rare now (though there are a couple in the APT VM).

BSG-75:
> Thanks, this is good stuff!
> 
> I think you should remove the logging call, and just replace it with a TODO comment like the others in the same file. Not because it's unwanted, but because we're integrating a proper logging solution soon (hopefully by the end of the month!), and `Debug.WriteLine` calls are somewhat rare now (though there are a couple in the APT VM).

Thanks for the comment! Removed `Debug.WriteLine`.

PR #219, opened 2018-12-12
Added a tooltip to display full filename / item name

BSG-75:
Now we are able to see full filename in any cases. Just hover the mouse on the item and it's full name will be displayed.
Effect:
image
image


tgjones:
Thank you!

PR #140, opened 2018-08-04
[WIP] Debug view + Pathfinder + Object movement

MarkusHiller:
*some debug informations (CTRL + D Collider + MousePos / CTRL + G GridPoints)
*Object movment with right click (initial version)
*GridSystem (Pathfinder) from map-data

image

PR #226, opened 2018-12-19
Debug overlay and debug drawing methods

paavohuhtala:
Based on parts of #140.

* Adds `DebugOverlay`, a 2D UI layer on top of `Scene3D`. The debug overlay can be toggled with F2. Right now it shows three debug strings in the top left corner:
  * "Screen": Cursor's 2D coords
  * "Terrain": The result of intersecting a ray from the cursor with the terrain
  * "Tile": The previous value divided by 10 and truncated to an integer 
  image
  * Text rendering is really slow right now, so enabling the overlay (which updates every frame) has a really high FPS impact.
* When the debug overlay is enabled, F3 can be used to toggle the rendering of collider outlines.
  image
  * These (like everything else in this PR) are rendered using software rendering. This could be rewritten to use GPU rendering.
* Adds a set of debug drawing methods to `DebugOverlay`, accessible from elsewhere via `game.Scene3D.DebugOverlay`. These include `AddLine`, `AddPoint`, `AddCoordAxes` and "immediate mode" versions of them prefixed with `Draw` instead of `Add`.


src/OpenSage.Game/Graphics/Cameras/Camera.cs
```diff
@@ -141,20 +141,27 @@ public Vector3 WorldToScreenPoint(Vector3 position)
...
-            return new Rectangle(
-                (int) (screenPosition.X - screenSize.Width / 2.0f),
-                (int) (screenPosition.Y - screenSize.Height / 2.0f),
+            return new RectangleF(
+                screenPosition.X - screenSize.Width / 2.0f,
+                screenPosition.Y - screenSize.Height / 2.0f,
                 screenSize.Width, screenSize.Height);
         }
 
+        internal bool IsWithinViewportDepth(Vector3 screenPosition)
```
tgjones: This should have “in”.

src/OpenSage.Game/Gui/DebugUi/DebugDrawable.cs
```diff
@@ -0,0 +1,96 @@
+﻿using System.Numerics;
+using OpenSage.Graphics.Cameras;
+using OpenSage.Mathematics;
+
+namespace OpenSage.Gui.DebugUi
```
tgjones: The namespace should be DebugUI (if I understand the current .NET. Among conventions correctly). I haven't always followed them myself, but we might as well use them for new code.

src/OpenSage.Game/Gui/DebugUi/DebugDrawable.cs
```diff
@@ -0,0 +1,96 @@
...
+        void Render(DrawingContext2D context, Camera camera);
+    }
+
+    public class DebugPoint : IDebugDrawable
+    {
+        public readonly Vector3 Position;
+        public readonly ColorRgbaF Color;
+        public float? Timer { get; set; }
+
+        public DebugPoint(Vector3 position, ColorRgbaF color, float? duration = null)
```
tgjones: These should have “in” modifiers.

src/OpenSage.Game/Gui/DebugUi/DebugDrawable.cs
```diff
@@ -0,0 +1,96 @@
...
+            }
+        }
+    }
+
+    public class DebugCoordAxes : IDebugDrawable
+    {
+        public readonly Vector3 Position;
+        public float? Timer { get; set; }
+
+        public DebugCoordAxes(Vector3 position, float? duration = null)
```
tgjones: This should have an “in” modifier.

src/OpenSage.Game/Gui/DebugUi/DebugDrawable.cs
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public class DebugLine : IDebugDrawable
+    {
+        public readonly Vector3 Start;
+        public readonly Vector3 End;
+        public readonly ColorRgbaF Color;
+        public float? Timer { get; set; }
+
+        public DebugLine(Vector3 start, Vector3 end, ColorRgbaF color, float? duration = null)
```
tgjones: These should have “in” modifiers.

src/OpenSage.Game/Gui/DebugUi/DebugDrawable.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        public void Render(DrawingContext2D context, Camera camera)
+        {
+            DebugDrawingUtils.DrawLine(context, camera, Start, End, Color);
+        }
+    }
+
+    internal static class DebugDrawingUtils
+    {
+        public static void DrawLine(DrawingContext2D context, Camera camera, Vector3 start, Vector3 end, ColorRgbaF color)
```
tgjones: These should have “in” modifiers.

src/OpenSage.Game/Gui/DebugUi/DebugOverlay.cs
```diff
@@ -0,0 +1,136 @@
...
+        public DebugOverlay(Scene3D scene3D, ContentManager contentManager)
+        {
+            _scene3D = scene3D;
+            _debugFont = contentManager.GetOrCreateFont("Arial", 16, FontWeight.Normal);
+            _debugStringBuilder = new StringBuilder();
+
+            _debugDrawables = new List<IDebugDrawable>();
+        }
+
+        public void AddPoint(Vector3 point, ColorRgbaF color, float? duration = null)
```
tgjones: All these methods should use “in” modifiers.

src/OpenSage.Game/Graphics/Cameras/Camera.cs
Qibbi: Probably not as numerics vectors are not read-only structs. One could use ref, it would look ugly, but not actually slow it down.

tgjones: I thought that because we're only accessing fields on `Vector3`, not methods, the compiler would be clever enough to not make a defensive copy. But that's only a guess.

Qibbi: Yes that's true, the question is if there will be a breaking change in the future.

src/OpenSage.Game/Gui/DebugUI/DebugOverlay.cs
```diff
@@ -32,32 +32,32 @@ public DebugOverlay(Scene3D scene3D, ContentManager contentManager)
             _debugDrawables = new List<IDebugDrawable>();
         }
 
-        public void AddPoint(Vector3 point, ColorRgbaF color, float? duration = null)
+        public void AddPoint(in Vector3 point, ColorRgbaF color, float? duration = null)
```
tgjones: Looks great - just these `ColorRgbaF` parameters should also have the `in` modifier, then it's ready to merge.

PR #228, opened 2018-12-20
Implement RenderItemCollection and multithreaded culling

paavohuhtala:
Refactors some bits of the rendering pipeline to make culling and sorting code easier to read and a bit faster. Implements parallel culling, which is enabled by default, though in my testing it hasn't yielded huge perf gains. Remains to be seen how it works on e.g CPUs with a low clockspeed and lots of cores.

Based on a few random benchmarks FPS improvement is about 5-10% on a 4-core CPU.


tgjones:
Nice improvement, thank you!

PR #231, opened 2018-12-27
Finished rotwk object ini parsing

Tarcontar:


src/OpenSage.Game/Data/Ini/Parser/IniParser.cs
```diff
@@ -320,7 +320,19 @@ public int[] ParseInLineIntegerArray()
...
-        private long ScanLong(IniToken token) => Convert.ToInt64(token.Text);
+        private long ScanLong(IniToken token)
+        {
+            try
+            {
+                return (long)Convert.ToUInt64(token.Text);
+            }
+            catch(Exception _)
+            {
+                if (token.Text.StartsWith("-"))
```
tgjones: What kind of input value causes this code to run?

src/OpenSage.Game/Data/Ini/Parser/IniParser.cs
```diff
@@ -711,8 +730,10 @@ private bool ParseIncludedFile<T>(T result, IIniFieldParserProvider<T> fieldPars
             while (includeFileName.StartsWith(".."))
             {
                 includeFileName = includeFileName.Remove(0, 3);
-                directory = directory.Substring(0, directory.LastIndexOf('\\'));
+                directory = directory.Substring(0, directory.LastIndexOf(Path.DirectorySeparatorChar));
             }
+            if (includeFileName.StartsWith(Path.DirectorySeparatorChar.ToString()))
+                includeFileName = includeFileName.Remove(0, 1);
```
tgjones: Please add braces here.

src/OpenSage.Game/Data/Ini/Parser/IniParser.cs
```diff
@@ -745,11 +766,13 @@ public void ParseFile()
...
                     }
 
-                    _dataContext.Defines.Add(macroName, macroExpansionToken.Value);
+                    _dataContext.Defines.Add(macroName.ToUpper(), macroExpansionToken.Value);
                 }
                 else if (fieldName == "#include")
                 {
                     var includeFileName = ParseQuotedString();
+                    if (includeFileName.StartsWith(Path.DirectorySeparatorChar.ToString()))
+                        includeFileName = includeFileName.Remove(0, 1);
```
tgjones: Add braces here.

src/OpenSage.Game/Data/Ini/Parser/IniParser.cs
Tarcontar: Sth like 999999999999999

Tarcontar:
Added braces

src/OpenSage.Game/Data/Ini/Parser/IniParser.cs
tgjones: Pretty sure this should be `Convert.ToInt64(token.Text)`, otherwise you're throwing away every negative number. I haven't yet been able to find an example of an out-of-range number like `999999999999999` (and I have installed the Edain mode). Can you post a snippet of an .ini file that has a value like that?

Tarcontar: image

Tarcontar: true, ill change it to `Convert.ToInt64(token.Text)` then

tgjones: I think the method implementation should be:

``` csharp
private long ScanLong(IniToken token)
{
    try
    {
        return Convert.ToInt64(token.Text);
    }
    catch (OverflowException)
    {
        return token.Text.StartsWith("-")
            ? long.MinValue
            : long.MaxValue;
    }
}
```

Qibbi: I'd suggest token.Text[0] == '-' as it's a lot faster and it's a given the text is more than 0 chars in length if an overflow occurred.

Qibbi: Also are you sure you want to use Int64 instead of Int32? I can't think of a value from the games which would actually be 64 bits.

Tarcontar: thats true, so should we remove long completely??

tgjones: We need ParseLong for a few things, but it doesn’t need to be used for new fields in this PR. I’m still not sure we’ve got the behaviour right. I assume Generals’ ini parser used sscanf, which seems to “crop” out-of-range values to “infinity” rather than crashing. Maybe we should try to do that.

Tarcontar: Well since this stuff only occurs in the edain mod and not in any of the vanilla games, i would guess the modders just wanted some kind of max value. So our current implementation is sufficient for that, no need to overthink it, its just for durations and ranges, where infinity would be overkill.

tgjones: I agree. Please could you change it to the implementation I posted above, then I can merge this?

Tarcontar: done

PR #233, opened 2018-12-29
[WIP] Construction prototype

paavohuhtala:


MustaphaTR:
I tested it a bit but rotation doesn't always seem to work properly most of the time. It placed buildings facing either unrelated facings or default.

> Ability to cancel with right click

I kinda feel like this could be implemented here.

paavohuhtala:
@MustaphaTR Note that buildings are rotated by dragging your mouse in the X-axis. There's no visual indicator for it yet and there's no deadzone, so even the slightliest movement after pushing the mouse button down changes the placement angle. If you're really careful you can place buildings in the default orientation.

MustaphaTR:
> Note that buildings are rotated by dragging your mouse in the X-axis.

Ok, then. I was thinking it is done like comparing first position to last, so it is gonna face the location where you take off the mouse button.

src/OpenSage.Game/Logic/OrderGeneratorSystem.cs
```diff
@@ -0,0 +1,91 @@
...
+        public void OnActivate()
+        {
+            if (!_worldPosition.HasValue || ActiveGenerator == null)
+            {
+                return;
+            }
+
+            var result = ActiveGenerator.TryActivate(Game.Scene3D);
+
+            if (result is OrderGeneratorResult.Success success)
```
tgjones: Should this be a “switch” statement?

src/OpenSage.Game/Logic/Orders/Order.cs
```diff
@@ -92,12 +92,12 @@ public override string ToString()
...
         }
 
-        public static Order CreateClearSelection(uint playerId)
+        public static Order CreateClearSelection(int playerId)
         {
             return new Order(playerId, OrderType.ClearSelection);
         }
 
-        public static Order CreateSetSelection(uint playedId, uint objectId)
+        public static Order CreateSetSelection(int playedId, uint objectId)
```
tgjones: Typo - playerId.

PR #236, opened 2019-01-01
DFloat/DVector3 and Quadtree

paavohuhtala:
* Adds deterministic numeric types, namely `DFloat` and `DVector3`. The implementation is based on [SoftFloat](https://github.com/CodesInChaos/SoftFloat) with a few applied refactorings and changes. I'm still not 100% sure if this is the right solution, but at least we have something to experiment with.
  * Since SoftFloat is binary compatible with IEEE754, we could potentially use SSE intrinsics for accelerating it, though I'm still not sure if that's safe or not. I wrote a prototype implementation of SSE intrinsics-based add operator for `Vector3`. It was removed in 2ee87e04ee8f9630a8a25f81f1b65e716ab7e677. 
  * The types are not yet used anywhere.
  * `SoftFloat` should be usable enough for basic maths, but it's missing a lot of functionality, namely division.
* Adds `Quadtree<T>`, an implementation of a [region quadtree](https://en.wikipedia.org/wiki/Quadtree#Region_quadtree).
  * It's not used anywhere yet, but it should be easy to use for accelerating collision detection (`ConstructBuildingOrderGenerator`) and culling (for both rendering and object picking).
  * We need to consider how to integrate the quadtree with `Scene3D` and how to keep it in sync. The quadtree need both the old and new bounding box boundaries of an object to update its position. The implementation should be fast enough to just regenerate the tree every logic update, but that seems wasteful.


src/OpenSage.Game/Mathematics/DVector3.cs
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        public override string ToString()
+        {
+            return $"({X}, {Y}, {Z})";
+        }
+
+        public override int GetHashCode()
+        {
+            var hashCode = -307843816;
```
tgjones: You can use Hash.Combine here.

src/OpenSage.Game/Mathematics/RectangleF.cs
```diff
@@ -142,18 +143,76 @@ public bool Contains(in Point2D point)
...
+            }
+
+            return obj is RectangleF other && Equals(other);
+        }
+
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                var hashCode = X.GetHashCode();
```
tgjones: Also here.

PR #245, opened 2019-01-04
Use the skirmish menu player settings when starting a game

feliwir:


src/OpenSage.Mods.Generals/Gui/ReplayMenuCallbacks.cs
```diff
@@ -71,10 +72,17 @@ public static void ReplayMenuSystem(Control control, WndWindowMessage message, C
 
                             context.Game.Scene2D.WndWindowManager.PopWindow();
 
+                            //TODO: set the correct factions & colors
```
paavohuhtala: Nitpickiest nitpick: could you add a space after `//` if you don't mind? 😅

PR #246, opened 2019-01-04
Fix animations with multi-bone-skinning

Tarcontar:


src/OpenSage.Game/Graphics/MeshVertex.cs
```diff
@@ -27,10 +32,15 @@ public struct TexCoords
         {
             new VertexLayoutDescription(
                 new VertexElementDescription("POSITION", VertexElementSemantic.TextureCoordinate, VertexElementFormat.Float3),
+                new VertexElementDescription("POSITION", VertexElementSemantic.TextureCoordinate, VertexElementFormat.Float3),
+                new VertexElementDescription("NORMAL", VertexElementSemantic.TextureCoordinate, VertexElementFormat.Float3),
                 new VertexElementDescription("NORMAL", VertexElementSemantic.TextureCoordinate, VertexElementFormat.Float3),
```
Tarcontar: Not sure why this is all VertexElementSemantic.TextureCoordinate, shouldnt these be "Position" or sth?

tgjones: It’s correct as it is. Veldrid.SPIRV always maps the semantics as TEXCOORDn.

src/OpenSage.Game/Content/ModelLoader.cs
```diff
@@ -319,17 +319,32 @@ private static FixedFunction.VertexMaterial[] CreateMaterials(W3dMesh w3dMesh)
             {
                 vertices[i] = new MeshVertex.Basic
                 {
-                    Position = w3dMesh.Vertices.Items[i],
-                    Normal = w3dMesh.Normals.Items[i],
+                    Position0 = w3dMesh.Vertices.Items[i],
+                    Position1 = w3dMesh.Vertices2 != null
```
tgjones: Is this necessary, or can you fallback to Vector3.Zero? Or put another way, if there are two bone influences, is Positions2 always present?

src/OpenSage.Game/Graphics/MeshVertex.cs
```diff
@@ -9,11 +9,16 @@ public static class MeshVertex
         [StructLayout(LayoutKind.Sequential)]
         public struct Basic
         {
-            public Vector3 Position;
-            public Vector3 Normal;
+            public Vector3 Position0;
```
tgjones: I think there are now enough skinning-specific attributes that we should have separate structs and code paths for skinning / non-skinning. But that can be done later.

Qibbi: Skinned meshes in at least TW has up to two bones, so wouldn't it make sense to have this a bit more dynamic and create the VertexLayoutDescription according to what is actually there to reduce the array size?

tgjones: I think we should do that, at the same time as we allow different vertex attributes for skinning / non-skinning.

tgjones: The problem will be in keeping the number of shader permutations under control. We don’t have a good way yet of dealing with those.

Qibbi: Doesn't really matter, does it? The shader uses registers which is what the VertexLayoutDescription is for.

tgjones: I assumed there would be warnings or errors if the vertex data doesn’t contain everything expected by the shader. At least, that’s how it works in D3D.

tgjones: We’d need to be able to basically compile multiple entry points per shader, one for each unique set of vertex attributes, and we can’t do that yet. It’s enough work that we will tackle it separately.

Qibbi: I see, TW uses macros and shader arrays for that.

src/OpenSage.Game/Graphics/MeshVertex.cs
Tarcontar: ok, but then the other enum values make no sense...

tgjones: Veldrid can be used without spirv-cross or Veldrid.Spirv. Those enum values can be used, just not by us.

src/OpenSage.Game/Content/ModelLoader.cs
Qibbi: Yes the second position is always there as it's the same vertex but in the other bone's space.

Tarcontar: well how would you test that for every model, need to iterate over every model and over every vertex influence?
Also what is the problem to use the default vertices here?

Qibbi: The way they apply the bone influences wouldn't even work without a corresponding set of vertices.

Qibbi: `		int index = blendIndices.x * 2;
		WorldPosition = BoneTransformPosition(WorldBones[index], WorldBones[index+1], InSkin.Position0) * InSkin.BlendWeights.x;
		index = blendIndices.y * 2;
		WorldPosition += BoneTransformPosition(WorldBones[index], WorldBones[index+1], InSkin.Position1) * InSkin.BlendWeights.y;`

Tarcontar: yea looks like it, wrote some code to test it though, so if the second bone index is not 0, there are always vertices2 and normals2

PR #248, opened 2019-01-07
Fix crash for particle system bones

feliwir:
Prevents crashes when not FXParticleSystem are available

### DIFF
--- a/src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
+++ b/src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
@@ -172,10 +172,15 @@ namespace OpenSage.Logic.Object
             {
                 foreach (var particleSysBone in conditionState.ParticleSysBones)
                 {
-                    var particleSystemTemplate = _contentManager.IniDataContext.FXParticleSystems.First(x => x.Name == particleSysBone.ParticleSystem);
+                    var particleSystemTemplate = _contentManager.IniDataContext.FXParticleSystems.FirstOrDefault(x => x.Name == particleSysBone.ParticleSystem);
                     if (particleSystemTemplate == null)
                     {
-                        particleSystemTemplate = _contentManager.IniDataContext.ParticleSystems.First(x => x.Name == particleSysBone.ParticleSystem).ToFXParticleSystemTemplate();
+                        particleSystemTemplate = _contentManager.IniDataContext.ParticleSystems.FirstOrDefault(x => x.Name == particleSysBone.ParticleSystem).ToFXParticleSystemTemplate();
+
+                        if (particleSystemTemplate == null)
+                        {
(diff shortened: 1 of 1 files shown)


src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
```diff
@@ -172,10 +172,15 @@ private W3dModelDrawConditionState CreateModelDrawConditionStateInstance(ModelCo
...
-                    var particleSystemTemplate = _contentManager.IniDataContext.FXParticleSystems.First(x => x.Name == particleSysBone.ParticleSystem);
+                    var particleSystemTemplate = _contentManager.IniDataContext.FXParticleSystems.FirstOrDefault(x => x.Name == particleSysBone.ParticleSystem);
                     if (particleSystemTemplate == null)
                     {
-                        particleSystemTemplate = _contentManager.IniDataContext.ParticleSystems.First(x => x.Name == particleSysBone.ParticleSystem).ToFXParticleSystemTemplate();
+                        particleSystemTemplate = _contentManager.IniDataContext.ParticleSystems.FirstOrDefault(x => x.Name == particleSysBone.ParticleSystem).ToFXParticleSystemTemplate();
+
+                        if (particleSystemTemplate == null)
+                        {
+                            throw new InvalidOperationException("Missing referenced particle system: " + particleSysBone.ParticleSystem);
```
paavohuhtala: This is fine for now, although we might want to convert this into a warning when we get around implementing a logging system.

src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
```diff
@@ -172,10 +172,15 @@ private W3dModelDrawConditionState CreateModelDrawConditionStateInstance(ModelCo
...
                 foreach (var particleSysBone in conditionState.ParticleSysBones)
                 {
-                    var particleSystemTemplate = _contentManager.IniDataContext.FXParticleSystems.First(x => x.Name == particleSysBone.ParticleSystem);
+                    var particleSystemTemplate = _contentManager.IniDataContext.FXParticleSystems.FirstOrDefault(x => x.Name == particleSysBone.ParticleSystem);
                     if (particleSystemTemplate == null)
                     {
-                        particleSystemTemplate = _contentManager.IniDataContext.ParticleSystems.First(x => x.Name == particleSysBone.ParticleSystem).ToFXParticleSystemTemplate();
+                        particleSystemTemplate = _contentManager.IniDataContext.ParticleSystems.FirstOrDefault(x => x.Name == particleSysBone.ParticleSystem).ToFXParticleSystemTemplate();
+
+                        if (particleSystemTemplate == null)
```
tgjones: This will never be reached, because it will crash on the line above when calling the ToFXParticleSystemTemplate method, if the result of FirstOrDefault is null.

PR #254, opened 2019-01-11
[WIP] parse missing stuff from bfme2 inis

Tarcontar:


src/OpenSage.Game/Data/Ini/CommandButton.cs
```diff
@@ -622,5 +697,192 @@ public enum CommandButtonRadiusCursorType
 
         [IniEnum("DOMINATE"), AddedIn(SageGame.Bfme)]
         Dominate,
+
+        //should this realy be enum values? or just names/strings
+        [IniEnum("ScatterAreaRadiusCursor"), AddedIn(SageGame.Bfme2)]
+        ScatterAreaRadiusCursor,
+
```
Tarcontar: Not sure if this Enum should be just a string value

src/OpenSage.Game.Tests/Data/Ini/IniFileTests.cs
```diff
@@ -25,38 +26,161 @@ public void CanReadIniFiles()
...
+                GameDefinition.FromGame(SageGame.Bfme2),
+                GameDefinition.FromGame(SageGame.Bfme2Rotwk),
             };
 
             foreach(var gameDefinition in gameDefinitions)
             {
-                InstalledFilesTestData.ReadFiles(".ini", _output, new List<IGameDefinition> { gameDefinition }, entry =>
+                var rootDirectories = InstallationLocators.FindAllInstallations(gameDefinition)
+                .Select(i => i.Path)
+                .Where(x => Directory.Exists(x))
```
tgjones: I believe both installation locators already check whether the path exists, so this check isn't necessary.

src/OpenSage.Game.Tests/Data/Ini/IniFileTests.cs
```diff
@@ -25,38 +26,161 @@ public void CanReadIniFiles()
...
+                            }
 
-                    dataContext.LoadIniFile(entry);
-                });
+                            switch (gameDefinition.Game)
+                            {
+                                case SageGame.Bfme2:
+                                    switch (filename)
+                                    {
+                                        case "ingamenotificationbox.ini":
```
tgjones: These are the lists of files still to do?

src/OpenSage.Game/Data/Ini/CommandButton.cs
tgjones: Pretty sure it was an enum, which has caused modders a lot of problems because it can't be extended. Let's keep it as an enum, but hopefully we can lift that restriction later.

Tarcontar: ok thx

src/OpenSage.Game.Tests/Data/Ini/IniFileTests.cs
Tarcontar: yes, will add a comment there

src/OpenSage.Game.Tests/Data/Ini/IniFileTests.cs
Tarcontar: removed

PR #255, opened 2019-01-11
Replace Effect and EffectMaterial abstraction with something better

tgjones:
The current `Effect` framework accounts for quite a chunk of time, if you look at OpenSAGE with a profiler. It could be much faster.


tgjones:
The overall themes are:

* Use "raw" Veldrid types like `Pipeline` and `ResourceSet`. Veldrid encourages grouping resources that update at the same frequency into resource sets, and this wasn't really possible with the existing `Effect` and `EffectParameter` abstractions.
* Explicitly create `ResourceLayout`s instead of loading them from reflected shader JSON files. We need to know about resource layouts in the code anyway, in order to group things into resource sets, so reflection information from shader JSON files didn't give us much.
* Add a set of `*ShaderResources` classes, to cache `Pipeline` and `ResourceLayout` objects for each shader set. Pipelines are created in different ways for each shader, so the logic for this is different for each `*ShaderResources` class.

src/OpenSage.Game/Content/MapLoader.cs
```diff
@@ -98,26 +89,49 @@ Texture LoadTexture(string name)
...
+            var terrainPatches = CreatePatches(
+                contentManager.GraphicsDevice,
+                heightMap,
+                mapFile.BlendTileData,
+                indexBufferCache,
+                materialResourceSet);
+
+            var cloudTexture = LoadTexture(mapFile.EnvironmentData?.CloudTexture ?? "tscloudmed.dds");
+
+            var cloudResoureLayout = AddDisposable(contentManager.GraphicsDevice.ResourceFactory.CreateResourceLayout(
```
paavohuhtala:
```suggestion
            var cloudResourceLayout = AddDisposable(contentManager.GraphicsDevice.ResourceFactory.CreateResourceLayout(
```

src/OpenSage.Game/Data/StreamFS/AssetReaders/W3dMeshReader.cs
```diff
@@ -1,14 +1,5 @@
...
-using OpenSage.Data.W3d;
+﻿using System.IO;
 using OpenSage.Data.W3x;
-using OpenSage.Graphics;
-using OpenSage.Graphics.Effects;
-using OpenSage.Graphics.Rendering;
-using OpenSage.Mathematics;
-using Veldrid;
 
 namespace OpenSage.Data.StreamFS.AssetReaders
```
paavohuhtala: Should the commented out and outdated code be removed from this file?

src/OpenSage.Game/Graphics/ModelMesh.cs
```diff
@@ -1,8 +1,10 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Numerics;
+using OpenSage.Content;
```
paavohuhtala:
```suggestion

```

src/OpenSage.Game/Graphics/Shaders/FixedFunctionShaderResources.cs
```diff
@@ -0,0 +1,279 @@
...
+                    DepthComparison,
+                    BlendEnabled,
+                    SourceFactor,
+                    DestinationColorFactor,
+                    DestinationAlphaFactor);
+            }
+
+            public override bool Equals(object obj)
+            {
+                return obj is PipelineKey && Equals((PipelineKey) obj);
```
paavohuhtala:
```suggestion
                return obj is PipelineKey a && Equals(a);
```

src/OpenSage.Game/Graphics/Shaders/ShaderSet.cs
```diff
@@ -0,0 +1,88 @@
+﻿using OpenSage.Utilities.Extensions;
+using Veldrid;
+using Veldrid.SPIRV;
+
+namespace OpenSage.Graphics.Shaders
+{
+    public sealed class ShaderSet : DisposableBase
+    {
+        private static byte _nextId = 0;
```
paavohuhtala:
```suggestion
        private static byte _nextId;
```

PR #256, opened 2019-01-12
Better game loop

### LINKED ISSUE #102: Implement proper game loop
At the time of writing the game loop is very simple: `Game.Tick()` is called as often as possible. `Game.Tick()` polls for input, processes incoming messages from the network message buffer, updates game systems and renders the frame. This has been fine for now, but it doesn't match how SAGE works.

As far as I know, SAGE works kinda like this:

* The scripting system [updates at 30 FPS](https://opensage.readthedocs.io/systems/scripting/index.html).
* Other game systems also update at 30 FPS.
* The game normally renders at ~30 FPS (GeForce Experience reports the framerate in the shell map as 32 fps)
* Network messages and (possibly) player input are processed at 15 FPS
  * Could we poll input more often to reduce input lag without breaking compatibility?

If I had to guess I'd say the original implementation runs a single game loop at 30 FPS, updates network messages and input every other frame and then does everything else, including rendering.

There are some special cases, some of which we need to support and some which we might want to support:

* When the pause menu is open, all units freeze, particles stop animating and vertex animations (tree sway) stop. At least in Generals, animated textures (rivers) continue animating, although this could be more of an oversight rather than a decision. The UI system operates normally.
* The "freeze time" scripting action works very similarly to the pause menu. The script which starts the pause should probably continue executing, so that the game doesn't get stuck. This requires further research.
* In single player skirmish Generals has a game speed slider, which goes from 15 to 60 + unlimited (shown as two minus signs) and defaults to 30. This affects both gameplay systems and rendering. At unlimited the game runs as fast as it can (on my computer about 700 FPS in a small 1v1 map).
  * On my computer with game speed set to 60 the game renders at 67 FPS, which is probably unintentional.

Personally I would love to see rendering supported at arbitrary framerates. This is probably the biggest enhancement we can do to for the original games, since AFAIK every single SAGE game is locked to 30 FPS. In the age of powerful enough hardware and enthusiasts having 120-240hz screens, it doesn't make sense to stick to 30.

paavohuhtala:
(Partially) resolves  #102.

* Splits updates into 2 kinds:
  * `LogicTick`
    * Run at 5Hz
    * Used for networking, order processing, unit movement etc.
  * `LocalLogicTick`
    * Runs on every frame
    * Should be used for everything else, including input handling, UI updates, animation, interpolation etc. 
* Splits timers into 2 kinds:
  * `MapTime` (name can be changed)
    * Only runs when logic ticks are running; relative to the start of the current map (hence the name)
    * Reset to 0 when the map changes 
    * Should be used for timing things that should pause when the game is paused (interpolation, animation, particles, camera animation)
  * `RenderTime`
    * Runs continuously and never resets.
    * Used for things that things that update even when the game is paused: Imgui, rendering, tracing/logging
* Map scripts are now ran at 30hz. We'll have to see how that works with synchronisation.
* `GameTimer` is now more precise, as it measures time using stopwatch ticks (instead of timespan ticks).
* `GameTime` is now passed by an `in` reference.
* Implement pause and step buttons.
image

### DIFF
--- a/src/OpenSage.Game/Diagnostics/GameLoopView.cs
+++ b/src/OpenSage.Game/Diagnostics/GameLoopView.cs
@@ -1,0 +1,22 @@
+﻿using ImGuiNET;
+
+namespace OpenSage.Diagnostics
+{
+    internal class GameLoopView : DiagnosticView
+    {
+        public GameLoopView(DiagnosticViewContext context) : base(context)
+        {
+
+        }
+
+        public override string DisplayName => "Game loop";
+
+        protected override void DrawOverride(ref bool isGameViewFocused)
+        {
+            ImGui.Text($"Logic frame: {Game.CurrentFrame}");
+            ImGui.Text($"Total time : {Game.UpdateTime.TotalGameTime}");
+            ImGui.Text($"Frame time : {Game.UpdateTime.ElapsedGameTime}");
+            ImGui.Text($"Cumulative update time error: {Game.CumulativeLogicUpdateError}");
+        }
+    }
+}
--- a/src/OpenSage.Game/Diagnostics/GameLoopView.cs
+++ b/src/OpenSage.Game/Diagnostics/GameLoopView.cs
@@ -4,8 +4,6 @@ namespace OpenSage.Diagnostics
 {
     internal class GameLoopView : DiagnosticView
     {
-        private ulong? breakOnFrame;
-
         public GameLoopView(DiagnosticViewContext context) : base(context)
         {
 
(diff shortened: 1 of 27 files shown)


tgjones:
This is looking great. What else do you want to do here?

src/OpenSage.Game/Diagnostics/GameLoopView.cs
```diff
@@ -0,0 +1,24 @@
+﻿using ImGuiNET;
+
+namespace OpenSage.Diagnostics
+{
+    internal class GameLoopView : DiagnosticView
```
tgjones: Please could you add “sealed” here? There’s supposed to be a potential speed up when the JIT knows a method can’t be overridden anymore.

src/OpenSage.Game/Diagnostics/GameLoopView.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        }
+
+        public override string DisplayName => "Game loop";
+
+        protected override void DrawOverride(ref bool isGameViewFocused)
+        {
+            ImGui.Text($"Logic frame: {Game.CurrentFrame}");
+            ImGui.Separator();
+            ImGui.Text($"Map time:    {Game.MapTime.TotalGameTime}");
```
tgjones: Shall we round these values to, say, 2 decimal places? I find it a bit hard to read at a glance.

src/OpenSage.Game/Utilities/TimeSpanUtility.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace OpenSage.Utilities
+{
+    public static class TimeSpanUtility
+    {
+        public static TimeSpan Max(TimeSpan a, TimeSpan b)
```
tgjones: Should this have “in”? I can’t remember how big a TimeSpan is.

paavohuhtala: `TimeSpan` is just a wrapper around `long`, so 64 bits.

https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/TimeSpan.cs#L68

PR #257, opened 2019-01-13
Fix BFME mainmenu

feliwir:


src/OpenSage.Game/Gui/Apt/ActionScript/VM.cs
```diff
@@ -85,7 +91,7 @@ public Value Execute(Function func, Value[] args, ObjectContext scope)
                 instr = stream.GetInstruction();
             }
 
-            return Value.Undefined();
+            return null;
```
tgjones: Why return null here? What does this mean?

feliwir: Functions can return a value when using the `Return` Instruction. Before i used the `Value.Undefined` to signal that there is no return code. But it seems like `Value.Undefined` is a valid return, so i changed this to be null. See this change here: https://github.com/OpenSAGE/OpenSAGE/pull/257/files#diff-af6541dbbcfe678870455d19e4e930b6R32

tgjones: Right, makes sense.

tgjones:
Thanks for the fix!

PR #261, opened 2019-01-15
Fix maps on newer SAGE games

feliwir:


src/OpenSage.Game/Content/SubsystemLoader.cs
```diff
@@ -110,6 +110,7 @@ private IEnumerable<string> GetSubsystemEntryName(Subsystem subsystem)
                         case SageGame.Bfme:
                         case SageGame.Bfme2:
                         case SageGame.Bfme2Rotwk:
+                            LoadFiles(@"Maps\MapCache.ini");
```
paavohuhtala: Could you move this logic to `GetFilesForSubsystem` or `Load` instead? It doesn't feel right that a side-effect free method like this one should open files and parse INIs.

src/OpenSage.Game/Content/MapLoader.cs
```diff
@@ -400,7 +403,8 @@ private static GameObject CreateGameObject(MapObject mapObject, Team[] teams, Co
                         
                         break;
 
-                    default:
+                    case RoadType.Start:
```
paavohuhtala: Is this just a syntactical clarification, or does it change behaviour in some way?

feliwir: It's a syntactical clarification. Since there are still 2 unknown chunks i wanted to use the chunk names explicitly

src/OpenSage.Game/Content/SubsystemLoader.cs
feliwir: Yup, will do

src/OpenSage.Game/Diagnostics/MainView.cs
```diff
@@ -61,15 +61,19 @@ public void Draw(ref bool isGameViewFocused)
                     {
                         foreach (var mapCache in _context.Game.ContentManager.IniDataContext.MapCaches)
                         {
-                            if (ImGui.MenuItem($"{mapCache.DisplayName} ({mapCache.Name})"))
+                            if (mapCache.IsMultiplayer && ImGui.MenuItem($"{mapCache.DisplayName} ({mapCache.Name})"))
```
tgjones: I want all maps to be visible in this menu, otherwise there's not (currently) any way to get to single player maps. If it crashes, we should fix the crash rather than hiding them altogether.

src/OpenSage.Game/Content/SubsystemLoader.cs
```diff
@@ -190,7 +190,15 @@ private IEnumerable<string> GetSubsystemEntryName(Subsystem subsystem)
                     throw new ArgumentOutOfRangeException(nameof(subsystem), subsystem, null);
             }
         }
-        
+
+        private void LoadFiles(params string[] files)
```
tgjones: Remove this unused method.

PR #268, opened 2019-01-20
Add initial support for trace output

tgjones:
This adds support for optionally outputting trace data to a .json file, which can then loaded into Chrome's profiler at chrome://tracing. This PR just adds very basic support, which needs to be fleshed out with more tracing calls.


tgjones:
One thing I'm not happy about with these changes is that, in every method where I add a tracing call, it indents the contents of the method. That results in a lot of changes... but I'm not sure if there's a better way to do it.

tgjones:
@feliwir I've fixed the crash when tracing is disabled.

PR #275, opened 2019-01-21
Implement more wnd callbacks (single player menu, exit in-game)

tgjones:


src/OpenSage.Game/Content/ModelLoader.cs
```diff
@@ -105,7 +105,7 @@ protected override Model LoadEntry(FileSystemEntry entry, ContentManager content
                     subObjects.Add(new ModelSubObject(w3dSubObject.Name, bone, mesh));
                 }
             }
-            else
+            else if (w3dMeshes.Count > 0)
```
feliwir: I guess this fixes a crash?

src/OpenSage.Game/Game.cs
```diff
@@ -423,11 +435,56 @@ public void StartGame(string mapFileName, IConnection connection, PlayerSetting[
...
+                0,
+                false);
+        }
+
+        public void StartMultiPlayerGame(
+            string mapFileName,
+            IConnection connection,
+            PlayerSetting[] playerSettings,
+            int localPlayerIndex)
+        {
```
feliwir: This will change in the future?

src/OpenSage.Game/Content/ModelLoader.cs
tgjones: Yes, the one you found in CHI01. Now that we can load it from the menu, it seemed somewhat related :)

src/OpenSage.Game/Game.cs
tgjones: It will need a lot more code than what's there now, if that's what you mean?

PR #279, opened 2019-01-26
Parse some missing ini stuff from ROTWK

Tarcontar:


src/OpenSage.Game/Gui/Apt/AptWindow.cs
```diff
@@ -74,10 +74,10 @@ internal void Render(DrawingContext2D drawingContext)
         {
             var fullsizeRect = new Rectangle(Point2D.Zero, _destinationSize);
 
-            if (BackgroundImage != null)
-            {
-                drawingContext.DrawImage(BackgroundImage.Texture, BackgroundImage.SourceRect, fullsizeRect);
-            }
+            //if (BackgroundImage != null)
```
tgjones: Is this intentional? If it’s the correct change, it should be deleted instead of commented out. cc @feliwir

Tarcontar: so i remove it for now?

PR #289, opened 2019-02-21
translation hotfix

Tarcontar:
- all my english texts where "missing" since the language was set to german, but all translation strings where in english.big, maybe investigate that further


Qibbi:
str doesn't have a way to set a language explicitly, which is why I have a parameter for that there but it defaults to en-us currently.
https://github.com/OpenSAGE/OpenSAGE/blob/d52cf66acfd3d943eadf37dc9cb67828a9e80be1/src/OpenSage.Game/Content/Translation/TranslationManager.cs#L219

OmniBlade:
The original game doesn't care about the language flagged in the csf and only loads the str or csf from the localised path based on the language set in the registry. You can actually install multiple languages side by side and twiddle the registry to change the language with the original, though you need to restart for the change to take effect. If no language is set in the registry I believe it falls back to the english path.

Qibbi:
I know, this was just a modified version of something I use myself for another project which does care about the set language.

PR #313, opened 2019-08-05
Make displaylist being sorted

feliwir:
I already did this change a while ago, but forgot to do a PR. The issue here was that objects in apt get rendered / handled in the wrong order since i though Dictionaries are sorted by default. Thankfully @Gwanw pointed that issue out to me.

### DIFF
--- a/src/OpenSage.Game/Gui/Apt/DisplayList.cs
+++ b/src/OpenSage.Game/Gui/Apt/DisplayList.cs
@@ -4,11 +4,11 @@ namespace OpenSage.Gui.Apt
 {
     public sealed class DisplayList
     {
-        public Dictionary<int, DisplayItem> Items { get; }
+        public SortedDictionary<int, DisplayItem> Items { get; }
 
         public DisplayList()
         {
-            Items = new Dictionary<int, DisplayItem>();
+            Items = new SortedDictionary<int, DisplayItem>();
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/OpenSage.Game/Gui/Apt/SpriteItem.cs
```diff
@@ -345,7 +346,7 @@ public override void RunActions(TimeInterval gt)
 
         public override bool HandleInput(Point2D mousePos, bool mouseDown)
         {
-            foreach (var item in Content.Items.Values)
+            foreach (var item in Content.Items.Values.Reverse())
```
feliwir: Buttons most be handled from top to bottom objects (in terms of depth)

src/OpenSage.Game/Gui/Apt/DisplayList.cs
```diff
@@ -4,11 +4,11 @@ namespace OpenSage.Gui.Apt
 {
     public sealed class DisplayList
     {
-        public Dictionary<int, DisplayItem> Items { get; }
+        public SortedDictionary<int, DisplayItem> Items { get; }
```
feliwir: DisplayItems are sorted by a depth value. Before they were just randomly in here (depending on the insertion order)

src/OpenSage.Game/Gui/Apt/SpriteItem.cs
tgjones: It would be better to store these in the correct order rather than reversing the order every time?

feliwir: The order is correct. The rendering happens exactly the opposite way, from far to near and happens much more frequently than the input handling :)

tgjones: And then can you store both directions, so it doesn’t need to be computed every frame?

feliwir: Yep, i’ll do that tomorrow

Tarcontar: Fixed

Issue #314, opened 2019-08-15
camera should start on starting position


DaanMeijer:
When a skirmish game is started, the view is at the center of the map. This means one has to search for their starting position.

feliwir:
Done by @DaanMeijer

PR #315, opened 2019-08-15
starts the game with the camera on the players starting building

DaanMeijer:
Fixes issue #314

### DIFF
--- a/src/OpenSage.Game/Game.cs
+++ b/src/OpenSage.Game/Game.cs
@@ -634,6 +634,11 @@ namespace OpenSage
 
                         players[i].SelectUnits(new [] { startingBuilding });
                     }
+
+                    if (players[i].IsHuman)
+                    {
+                        localPlayerIndex = i;
+                    }
                 }
 
                 Scene3D.SetPlayers(players, players[localPlayerIndex]);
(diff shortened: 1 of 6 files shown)


src/OpenSage.Game/Game.cs
```diff
@@ -578,6 +578,11 @@ public void ShowMainMenu()
 
                         players[i].SelectUnits(new [] { startingBuilding });
                     }
+
+                    if (players[i].IsHuman)
```
feliwir: We shouldn't set this inside of here. Replays have no local player

src/OpenSage.Game/Scene3D.cs
```diff
@@ -157,6 +157,9 @@ public void SetPlayers(IEnumerable<Player> players, Player localPlayer)
 
             LocalPlayer = localPlayer;
 
+            var MainUnit = LocalPlayer.SelectedUnits.First();
```
feliwir: Do the local players always have selected units at the start? A player isn't required to have starting units at all when configured like that inside the PlayerTemplate.ini. 
This would fail when the SelectedUnits count is 0, so please but an if statement arround the two lines.

src/OpenSage.Game/Game.cs
tgjones: We already set the local player index in this method, so I think this is okay. This change makes it a little better, but you're right that ultimately this isn't the right place to do it.

PR #323, opened 2019-08-20
show/hide muzzle bones based on firing state

DaanMeijer:


src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
```diff
@@ -119,6 +119,20 @@ public override void UpdateConditionState(BitArray<ModelConditionFlag> flags)
             }
 
             SetActiveConditionState(bestConditionState);
+
+            bestConditionState.WeaponMuzzleFlashes.ForEach(x => {
+                var visible = flags.Get(ModelConditionFlag.FiringA);
```
Qibbi: This is wrong. FiringA is used when firing from the primary weapon slot. FiringB from the secondary etc. So you'd have to check that.
I would also recommend to move this out of this function. Maybe make it event driven and let the muzzle flash object subscribe to it when needed.
Also usually the muzzle flash is usually only appearing in the default, as the others are basically overrides. So you should first check the current state and if not found check the default.

Another thing I noticed (which is not related to your code) is that EA's SAGE games choose the current condition state in quite a different manner. They pick the first state which has all the bits set, regardless if there is another more relevant state coming after that. It might not be as intuitive but should offer better performance.

src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -205,8 +205,12 @@ internal void LocalLogicTick(in TimeInterval gameTime, float tickT, HeightMap he
             {
                 drawModule.SetWorldMatrix(worldMatrix);
             }
+
+            CheckForDirtyModelConditionFlags();
         }
 
+
+
```
feliwir: Remove empty lines

src/OpenSage.Mods.Generals/Gui/GeneralsControlBar.cs
```diff
@@ -118,7 +118,7 @@ public void Update(Player player)
 
             _moneyDisplay.Text = $"$ {player.Money}";
 
-            if (player.SelectedUnits.Count > 0)
+            if (player.SelectedUnits.Count > 0 && player.SelectedUnits.First().Owner == player)
```
feliwir: I think that is already part of a different PR of you?

src/OpenSage.Game/Data/Ini/BitArray`1.cs
```diff
@@ -112,6 +112,12 @@ public override int GetHashCode()
             return HashCode.Combine(_inner.Count);
         }
 
+        public BitArray<TEnum> Clone()
```
feliwir: I think you could inherit from ICloneable

src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
```diff
@@ -119,6 +119,20 @@ public override void UpdateConditionState(BitArray<ModelConditionFlag> flags)
             }
 
             SetActiveConditionState(bestConditionState);
+
+            bestConditionState.WeaponMuzzleFlashes.ForEach(x => {
```
tgjones: Nitpick - please avoid lambdas / delegates in anything that happens in the main game loop. Lambdas cause memory allocation, and therefore garbage collection. This can be rewritten as a standard `foreach`.

src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
```diff
@@ -245,7 +287,7 @@ internal override void SetWorldMatrix(in Matrix4x4 worldMatrix)
 
     internal sealed class W3dModelDrawConditionState : DisposableBase
     {
-        private readonly ModelInstance _modelInstance;
+        public readonly ModelInstance _modelInstance;
```
tgjones: If you need to access this from the outside the class, please rename it to `ModelInstance`.

DaanMeijer: Was fixed by someone else adding a public Model

src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
DaanMeijer: Rewrote it using a foreach

src/OpenSage.Game/Data/Ini/BitArray`1.cs
DaanMeijer: Fixed

src/OpenSage.Mods.Generals/Gui/GeneralsControlBar.cs
DaanMeijer: You're right.

PR #326, opened 2019-09-06
Refactor MapLoader

tgjones:


src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -17,6 +18,61 @@ namespace OpenSage.Logic.Object
...
+            ContentManager contentManager,
+            GameObjectCollection parent,
+            in Vector3 position)
+        {
+            var gameObject = contentManager.InstantiateObject(mapObject.TypeName, parent);
+
+            // TODO: Is there any valid case where we'd want to return null instead of throwing an exception?
+            if (gameObject == null)
+            {
+                return null;
```
feliwir: Should we maybe log this? `InstantiateObject`doesn't do so itself

src/OpenSage.Game/Scene3D.cs
```diff
@@ -84,45 +90,204 @@ internal IEnumerable<AttachedParticleSystem> GetAllAttachedParticleSystems()
...
+
+                        break;
+
+                    case RoadType.Start:
+                    case RoadType.End:
+                        var roadEnd = mapObjects[++i];
+
+                        // Some maps have roads with invalid start- or endpoints.
+                        // We'll skip processing them altogether.
+                        if (mapObject.TypeName == "" || roadEnd.TypeName == "")
```
feliwir: Maybe `String.IsNullOrEmpty` might be a bit more readable? Or checking for `Length == 0`

src/OpenSage.Game/Terrain/Bridge.cs
```diff
@@ -18,36 +19,30 @@ public sealed class Bridge : DisposableBase
...
         {
-            _model = model;
-
-            _modelInstance = AddDisposable(_model.CreateInstance(contentManager));
-
-            _modelInstance.Update(TimeInterval.Zero);
+            var template = contentManager.IniDataContext.FindBridgeTemplate(mapObject.TypeName);
 
-            _towers = new List<GameObject>();
+            var modelPath = Path.Combine("Art", "W3D", template.BridgeModelName + ".W3D");
```
feliwir: This won't work for >=BFME since the modelpath is constructed differently there (models are in a subfolder named with the first two characters of the modelname)

PR #328, opened 2019-09-11
Refactor more content loading, including ini and models

tgjones:


src/OpenSage.Game/Data/Ini/LazyAssetReference.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+
+namespace OpenSage.Data.Ini
+{
+    public sealed class LazyAssetReference<T>
```
feliwir: This looks similar to `System.Lazy`. What are the differences?

PR #331, opened 2019-09-21
Fix scaling being applied in geometry view

feliwir:
It makes no sense to scale the geometry in when there is no parent AptWindow


feliwir:
> 
> 
> In what cases would there be no parent window?

It's for the `Apt geometry window` in our developer mode view. That particular mode crashed before since the `AptWindow` is null in that case. I think in the long run we want to decouple the rendering of shapes entirely from the Window, but that isn't done yet.

PR #338, opened 2019-09-29
Improve locomotor implementation

tgjones:


feliwir:
@tgjones Hey, what was missing on this PR? I just found that issue since BFME doesn't have support for Locomotors yet and i don't want to work on this while this is not done.

PR #345, opened 2019-12-22
Optimize BitArray and fix model condition state selection

paavohuhtala:
Based on #344


paavohuhtala:
We could do that, or we could use a normal array instead of storing the data in fields. It would enable arbitrary sized bit arrays, but it would also add an extra allocation and layer of indirection to each bit array. Let's tackle the issue when it becomes relevant.

PR #348, opened 2019-12-24
Fix duplicate GameObject insertion

Skleni:
`GameObjects` were added twice to the collection because `FromMapObject` already calls `parent.Add(mapObject.TypeName)`.


src/OpenSage.Game/Scene3D.cs
```diff
@@ -188,10 +188,6 @@ internal Scene3D(Game game, MapFile mapFile)
                                 position.Z += heightMap.GetHeight(position.X, position.Y);
 
                                 var gameObject = GameObject.FromMapObject(mapObject, teams, loadContext.AssetStore, gameObjects, position);
-                                if (gameObject != null)
```
feliwir: It doesn't make sense to keep the gameObject variable or does it?

Skleni: Right...I'll remove it.

Issue #341, opened 2019-12-14
Road rendering


Skleni:
Roads are currently rendered as straight lines, without connections between the segments. I'm currently trying to figure out how exactly the rendering works in SAGE. I'll document what I found out here.

* Roads are made up from segments
* The segments are stored in the .map files as a pair of objects (the start and end point) with the `RoadType` flag `Start`/`End`
* Road segments are connected if the have the same object type 
and share one point (each segments still has two points, the points are considered shared if they are located at the same coordinates)
* Each point can have additional `RoadType` flags that affect how two segments of the same type are connected
  * Angled
  * TightCurve
  * If none of these two is present, the default is BroadCurve

The direction of the edges (determined by the order of the two point objects) *should* not affect the rendering (there's no way to control or even see it in WB), but it does:

![road segment directions](https://user-images.githubusercontent.com/10420514/70854427-d0dbba00-1ebb-11ea-9e38-8305a9fab446.png)

I added waypoints as labels to show the flags of the start and end points and the directions of the edges. The two graphs are identical, except that the upper one is cyclic while in the lower one the left edge is reversed. This seems to cause the lower left corner to be rendered incorrectly as a broad curve, although both flags at that position are set to `TightCurve`. There seem to be *some* rules though:
* Setting the lower left corner to `Angled` does render an angled connection correctly .
* Setting the upper left corner to `TightCurve` renders the lower left corner as tight curve correctly.

Skleni:
Here are all the possible combinations of flags (*A*ngled, *T*ightCurve, *B*roadCurve) for three nodes that are connected the "wrong" way (X -> Y <- Z instead of X -> Y -> Z) and how they are rendered:

X | Y | Z | Result
-- | -- | -- | --
A | A | A | A
A | A | T | A
A | A | B | A
A | T | A | B
A | T | T | T
A | T | B | B
A | B | A | B
A | B | T | T
A | B | B | B  
T | A | A | A
T | A | T | A
T | A | B | A
T | T | A | B
T | T | T | T
T | T | B | B
T | B | A | B
T | B | T | T
T | B | B | B
B | A | A | A
B | A | T | A
B | A | B | A
B | T | A | B
B | T | T | T
B | T | B | B
B | B | A | B
B | B | T | T
B | B | B | B

Apparently, these are the rules:
```
if Y == Angled
    Angled
else if Z == TightCurve
    TightCurve
else
    BroadCurve
```

Skleni:
[...]

PR #350, opened 2019-12-28
Roads

Skleni:
This pull request improves the rendering of roads ingame. I refactored the ´RoadNetwork´ class to use segments instead of nodes and edges because some nodes (crossings) also need to be rendered with a texture. This also allows an optimization where only one vertex buffer per road network is created instead of one per road segment.

There is still quite a lot to do, but the current version can already draw connected roads:

grafik

- [ ] Find bug where roads are randomly drawn below the terrain
- [x] Create infrastructure for various road types (crossings, ...)
- [ ] Figure out texture atlas coordinates for all road types [WIP]
- [x] Implement straight roads
- [x] Implement angled connections between roads
- [ ] Implement broad curves [WIP]
- [ ] Implement tight curves
- [ ] Implement T crossings [WIP]
- [ ] Implement X crossings
- [ ] Implement symmetric Y crossings
- [ ] Implement asymmetric Y crossings 
- [ ] Improve rendering on uneven terrain
- [ ] Implement end caps/road joins

See [this issue](https://github.com/OpenSAGE/OpenSAGE/issues/341) for additional info on what I've found out so far.


src/OpenSage.Game/Terrain/HeightMap.cs
```diff
@@ -27,10 +27,17 @@ public float GetHeight(float x, float y)
...
             y = y / HorizontalScale + _heightMapData.BorderWidth;
 
-            if (x >= Width || y >= Height || x < 0 || y < 0)
-            {
-                return 0;
-            }
+            if (x < 0)
+                x = 0;
+
+            if (x >= Width)
```
feliwir: I think `Math.Clamp` would make this more readable: https://docs.microsoft.com/de-de/dotnet/api/system.math.clamp?view=netcore-3.0
e.g. `x = Math.Clamp(x, 0, Width -1);`

src/OpenSage.Game/Terrain/HeightMap.cs
```diff
@@ -27,10 +27,17 @@ public float GetHeight(float x, float y)
...
+                x = Width - 1;
+
+            if (y < 0)
```
feliwir: Same as above

tgjones:
Amazing stuff, thanks!

PR #351, opened 2019-12-29
Pathfinding

feliwir:
- [x] Implement A star algorithm
- [x] Incorporate impassability data from the map
- [x] Check water areas 
- [ ] Make gameobjects impassable
- [ ] Include bridges as passable areas


feliwir:
Water areas are marked as impassable automatically by the world builder

src/OpenSage.Game/Navigation/Graph.cs
```diff
@@ -0,0 +1,123 @@
...
+                        if (distanceToNext < next.ShortestKnownDistance)
+                        {
+                            next.ShortestKnownDistance = distanceToNext;
+                            _unexpandedNodes.UpdatePriority(next, estimatedDistanceToEnd);
+                            next.CameFrom = current;
+                        }
+                    }
+                    else
+                    {
+                        next.ShortestKnownDistance = distanceToNext;
```
Skleni: Is it possible that an already visited node's `CameFrom` points to `next`? In that case its `ShortestKnownDistance` would also have to be updated.

paavohuhtala: I'm not sure if this could happen. Correct me if I'm wrong but this shouldn't be an issue as long as we assume costs are symmetrical (going from A to B has the same cost as B to A).

Skleni: I'm not sure either :D I remember having this problem some time ago, but I couldn't reproduce it now.

PR #360, opened 2020-01-06
Don't crash for non existent filesystems

feliwir:
The **Load replays** button was crashing since it tried mounting the **Replays** folder which doesn't exist


feliwir:
> 
> 
> Approved in principle, though I wonder if we should either not create potentially invalid file systems in the first place or at least log a warning. I could see this potentially causing some hard to debug issues.

I agree here, but i didn't want to perform a File.Exists check right inside the gamelogic. I'll add a warning

PR #362, opened 2020-01-06
Curved roads

Skleni:
Adds support for tight and broad curves.

The texture mapping isn't perfect, but then again, neither is the orignal.

The decision which curve type is chosen is somewhat arbitrary (see #341), this algorithm isn't implemented yet. We only use tight and broad curves if both edges agree on the type.

grafik

### DIFF
--- a/src/OpenSage.Game/Terrain/Roads/RoadConstants.cs
+++ b/src/OpenSage.Game/Terrain/Roads/RoadConstants.cs
@@ -1,0 +1,11 @@
+﻿namespace OpenSage.Terrain.Roads
+{
+    public static class RoadConstants
+    {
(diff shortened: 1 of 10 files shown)


src/OpenSage.Game/Terrain/Roads/RoadConstants.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace OpenSage.Terrain.Roads
+{
+    public static class RoadConstants
+    {
+        // TODO figure out exact threshold
+        public const float MinCurveAngle = 0.47f;
```
feliwir: I guess they used some degree and converted it to radians. The next degree would be 27° with 0.47123889804.

PR #368, opened 2020-01-07
Render building preview when placing building

tgjones:


src/OpenSage.Game/Graphics/Shaders/MeshShaderResources.cs
```diff
@@ -155,7 +155,7 @@ public struct RenderItemConstantsPS
         {
             public Vector3 HouseColor;
 #pragma warning disable CS0169
```
feliwir: The warning can be removed

PR #378, opened 2020-01-11
Object prerequisites

tgjones:
Implements object prerequisites, so we can gray out - and disable - buttons for stuff that can't be built.

image


src/OpenSage.Game/Assets/Shaders/Sprite.frag
```diff
@@ -62,5 +62,11 @@ void main()
 
     textureColor *= in_Color;
 
+    if (_SpriteConstants.Grayscale)
+    {
+        float gray = 0.299 * textureColor.r + 0.587 * textureColor.g + 0.114 * textureColor.b;
```
feliwir: There are multiple formulas to convert color to grayscale, maybe we should document which one we use?

PR #379, opened 2020-01-11
Scripting

Skleni:
Refactored the scripting system so that all players' scripts are executed.
Refactored waypoints so that branches and loops in paths are possible.
Added a seedable `Random` instance to `Scene3D`

Added script actions:
MoveNamedUnitTo
NamedFollowWaypoints

### DIFF
--- a/src/OpenSage.Game/Scripting/ScriptingSystem.cs
+++ b/src/OpenSage.Game/Scripting/ScriptingSystem.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using OpenSage.Data.Map;
 
 namespace OpenSage.Scripting
@@ -24,6 +25,8 @@ namespace OpenSage.Scripting
 
     public sealed class ScriptingSystem : GameSystem
     {
+        public static NLog.Logger Logger => NLog.LogManager.GetCurrentClassLogger();
+
         // How many updates are performed per second?
         public const int TickRate = 30;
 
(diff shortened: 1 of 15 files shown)


src/OpenSage.Game/Scripting/ScriptingSystem.cs
```diff
@@ -65,34 +68,26 @@ internal override void OnSceneChanging()
 
         internal override void OnSceneChanged()
         {
-            _mapScripts = Game.Scene3D?.Scripts;
+            _playerScripts = Game.Scene3D?.PlayerScripts;
         }
 
-        public void EnableScript(string name)
+        public MapScript FindScript(string name)
         {
```
Tarcontar: this function returns always null?

src/OpenSage.Game/Scripting/ScriptingSystem.cs
```diff
@@ -65,34 +68,26 @@ internal override void OnSceneChanging()
...
-        {
-            var mapScript = _mapScripts?.FindScript(name);
-            if (mapScript != null)
+            for (int i = 0; i < _playerScripts.Length; i++)
             {
-                mapScript.IsActive = false;
+                var script = _playerScripts[i].FindScript(name);
+                if (script != null)
+                {
+                   return null;
```
Tarcontar: should return script here

Skleni: Good catch, thanks!

PR #382, opened 2020-01-12
Work on object upgrades

feliwir:
* Adds upgrades to the production queue
* Use **BuildTime** instead of **BuildCost**. This is still different from the original games production speed. We probably need to take of GameSpeed aswell.
* Each gameobject has a list of granted upgrades

### DIFF
--- a/src/OpenSage.Game/Logic/Object/UpgradeDefinition.cs
+++ b/src/OpenSage.Game/Logic/Object/UpgradeDefinition.cs
@@ -4,16 +4,16 @@ using OpenSage.Gui;
 
 namespace OpenSage.Logic.Object
 {
-    public sealed class Upgrade : BaseAsset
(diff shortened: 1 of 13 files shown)


src/OpenSage.Game/Logic/Object/UpgradeDefinition.cs
```diff
@@ -4,16 +4,16 @@
 
 namespace OpenSage.Logic.Object
 {
-    public sealed class Upgrade : BaseAsset
+    public sealed class UpgradeDefinition : BaseAsset
```
tgjones: The C&C3 SDK calls this `UpgradeTemplate` - can we use that name too please?

feliwir: Will do after my current game session is finished

PR #390, opened 2020-01-18
UserData MapCache.ini

Skleni:
Generate the MapCache.ini for user maps on startup. This needs to be done because the user maps could have been modified.

The CRC is still missing (as we don't know the exact algorithm yet) and so are the properties added after ZH.

It's still useful already because you don't have to start the original games any more to generate the file when a new map is added.

### DIFF
--- a/src/OpenSage.Game/Game.cs
+++ b/src/OpenSage.Game/Game.cs
@@ -7,6 +7,7 @@ using OpenSage.Audio;
 using OpenSage.Content;
 using OpenSage.Data;
 using OpenSage.Data.Apt;
+using OpenSage.Data.Ini;
 using OpenSage.Data.Map;
 using OpenSage.Data.Rep;
 using OpenSage.Data.Wnd;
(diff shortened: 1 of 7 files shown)


src/OpenSage.Game/Game.cs
```diff
@@ -402,12 +403,7 @@ public NetworkMessageBuffer NetworkMessageBuffer
                     _userDataFileSystem = AddDisposable(new FileSystem(FileSystem.NormalizeFilePath(UserDataFolder)));
                     ContentManager.UserDataFileSystem = _userDataFileSystem;
 
-                    // TODO: Re-generate MapCache.ini for user maps
-                    var file = _userDataFileSystem.GetFile(@"Maps\MapCache.ini");
-                    if (file != null)
-                    {
-                        ContentManager.LoadIniFile(file);
-                    }
+                    new UserMapCache(ContentManager).Initialize(AssetStore);
```
feliwir: Is it on purpose that you don't store the instance? If so, shouldn't this be static?

Skleni: It's not static because it stores the `ContentManager` in a field and uses it in multiple methods...but other than that, it's work is done after `Initialize`. Maybe `UserMapCacheInitializer` would be a better name?

feliwir: I think we can leave it as is if there's no other use for that instance. I was just confused since it's a bit unusual to see something like this :)

PR #397, opened 2020-01-23
auto game detection in folder 

MetaIdea:
- automatically detect the sage game when sitting in it's root folder (portability)
- optional startparameter for gamepath


src/OpenSage.Launcher/Program.cs
```diff
@@ -60,9 +63,67 @@ public static void Main(string[] args)
...
+            var DetectedGame = opts.Game;
+            var GameFolder = opts.GamePath;
+            var ForceUseEnvironmentGamePath = true;
 
-            var definition = GameDefinition.FromGame(opts.Game);
+            if (GameFolder == null)
+            {
+                GameFolder = Environment.CurrentDirectory;
+            }
+            if (File.Exists(Path.Combine(GameFolder, "generals.exe")) && File.Exists(Path.Combine(GameFolder, "INI.big")))
```
feliwir: I think it would make sense to put this in `IGameDefinition`, so each game definition has a **Probe** method returning a bool wether the path is the correct game installation folder. Then you could just iterate over all GameDefinition and remove this logic here

src/OpenSage.Mods.Bfme2/Bfme2Definition.cs
```diff
@@ -37,5 +37,7 @@ public OnDemandAssetLoadStrategy CreateAssetLoadStrategy()
         }
 
         public static Bfme2Definition Instance { get; } = new Bfme2Definition();
+
+        public string LauncherExecutable => "lotrbfme.exe";
```
Tarcontar, line 41: has to be "lotrbfme2.exe"

Tarcontar: done

src/OpenSage.Launcher/Program.cs
Tarcontar: done

PR #391, opened 2020-01-18
Fix crashes for BFME

feliwir:

PR #400, opened 2020-01-28
Fix some BFME II crashes

tgjones:
Alternative fixes for #391.

### DIFF
--- a/src/OpenSage.Game/Logic/Object/Body/HighlanderBody.cs
+++ b/src/OpenSage.Game/Logic/Object/Body/HighlanderBody.cs
@@ -1,7 +1,39 @@
 ﻿using OpenSage.Data.Ini;
+using OpenSage.Mathematics.FixedMath;
 
 namespace OpenSage.Logic.Object
 {
+    public class HighlanderBody : ActiveBody
+    {
+        private readonly GameObject _gameObject;
+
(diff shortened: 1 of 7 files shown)


src/OpenSage.Game/Logic/Object/Body/HighlanderBody.cs
```diff
@@ -1,7 +1,39 @@
 ﻿using OpenSage.Data.Ini;
+using OpenSage.Mathematics.FixedMath;
 
 namespace OpenSage.Logic.Object
 {
+    public class HighlanderBody : ActiveBody
+    {
+        private readonly GameObject _gameObject;
```
feliwir: As the CI noted, this never gets initialized

src/OpenSage.Game/Logic/Object/Body/HighlanderBody.cs
```diff
@@ -1,7 +1,39 @@
...
+                if (Health < (Fix64)0)
+                {
+                    Health = (Fix64) 0;
+                }
+
+                // TODO: DamageFX
+
+                if (Health <= Fix64.Zero)
+                {
+                    _gameObject.Die(deathType);
```
feliwir: This will probably crash as a result

tgjones: Thanks, fixed.

PR #402, opened 2020-02-02
Add targetting of objects

feliwir:
* Add attack sounds
* Add two new orders
* Add logic for targetting objects
* Prevent crashes when performing ForceAttack with units that can't attack


src/OpenSage.Game/Logic/Orders/OrderType.cs
```diff
@@ -10,7 +10,9 @@ public enum OrderType
         BuildObject = 1049,
         Sell = 1052,
         DrawBoxSelection = 1058,
-        AttackGround = 1061,
+        AttackObject = 1059,
+        ForceAttackObject = 1060,
```
feliwir: The game writes 2 different orders depending wether or not it is a force attack. So it makes sense to rename **AttackGround** aswell

PR #403, opened 2020-02-05
Fix armor crash

feliwir:


tgjones:
I think this needs to be rebased?

Issue #405, opened 2020-02-21
Memo: how fonts are handled in Sage Games


BSG-75:
These are taken from Discord discussion between me, @Qibbi and @feliwir on this issue: https://github.com/OpenSAGE/OpenSAGE/issues/398
I'm posting them there so it's not lost in the chat history
```
Language9x.ini ... it seems Generals uses a postfix for their ini names which denotes that it has another encoding... 
seems the 9x postfix is only for Language and HeaderTemplate
***9x.ini is used up to BFME, 
BFME2 changed that to only use fontsubstitution.ini, RA3 has that stuff inside compiled data

RA3+ just calls AddFontResourceExA (from GDI32.dll) on all fonts present in data/fonts/
TW I'm not 100%

Arial just has a few simple codepages, only Arial Unicode has the unicode stuff and asian languages
Microsoft used to ship it with Microsoft Office, but stopped shipping it in recent versions of Office.

That could be the reason why Generals uses Times New Roman, TNR seems to be the fallback fallback font, that one's hardcoded into the engine.
basically the "if we cannot find a font we actually want to do Times New Roman"

maybe we want something like Code2000
it also supports Tengwar and Klingon!
```

It might also be a good idea to wait for this one for handling fallback fonts:
https://github.com/SixLabors/Fonts/issues/117

Issue #398, opened 2020-01-28
OpenSAGE does not display some languages properlyXD


GenKyoko:
In the original CCG, the ttf font in big can be read normally, but XD cannot be displayed normally here....
![20200128204833](https://user-images.githubusercontent.com/54131467/73269194-54a4f800-4217-11ea-8953-6b318c1074c5.png)

BSG-75:
The issue is caused by OpenSage using "Arial" as the fallback font instead of `UnicodeFontName` specified in `data/(Language)/Language9x.ini`:
https://github.com/OpenSAGE/OpenSAGE/blob/1b74be7f5a2d031173da9e8bd09c3d3c71d5034c/src/OpenSage.Game/Content/FontManager.cs#L12

BSG-75:
@GenKyoko This should be fixed in https://github.com/OpenSAGE/OpenSAGE/pull/406
Can you try it?

GenKyoko:
![1TZ 01R}YI(OMN {B)~5~KY](https://user-images.githubusercontent.com/54131467/78953852-ba153f00-7b0c-11ea-8352-d2bc7172740b.png)
Well... still some problems

BSG-75:
> ![1TZ 01R}YI(OMN {B)5KY](https://user-images.githubusercontent.com/54131467/78953852-ba153f00-7b0c-11ea-8352-d2bc7172740b.png)
> Well... still some problems

Can you post a screenshot of OpenSage's console?

GenKyoko:
![NNRDS$}FW2I1SW6~CHLLN{2](https://user-images.githubusercontent.com/54131467/79093534-1daaa100-7d87-11ea-819c-333f87c07365.png)

BSG-75:
@GenKyoko Looks like I can't reproduce your issue, can you debug OpenSage by placing breakpoints in the following places, to check where the problem occured?

Here: (Check if `fileSystem.GetFile(...)`  has succeeded)
https://github.com/OpenSAGE/OpenSAGE/blob/eabd45692674a787557a2c4a1b1ab43b7615536d/src/OpenSage.Game/Content/Translation/TranslationManager.cs#L249

And here: (Check if a Traditional Chinese ("zh-Hant") `CultureInfo` has been successfully created)
[...]

PR #406, opened 2020-03-14
Partially address font issues

BSG-75:
Windows by itself can handle font fallback, but for OpenSage which uses cross platform libraries, it could be quite hard, especially when our libraries cannot handle non-English font family names yet.
This pull request will:
- [X] Allow reading those "9x" ini files (which contain stuffs in locale-specific encoding) mentioned in https://github.com/OpenSAGE/OpenSAGE/issues/405
- [X] Use a user-configurable JSON file (Which will be `UserSettingsFiles/FontFallbackSettings.json`) as a temporary solution to handle font fallbacks, which for now should fix Chinese characters rendering in https://github.com/OpenSAGE/OpenSAGE/issues/398


src/OpenSage.Game/Content/Translation/Providers/CsfTranslationProvider.cs
```diff
@@ -21,7 +21,7 @@ private enum LanguageGenerals
             [Display("ja-JP")] Japanese,
             [Display("en-US")] Jabber,
             [Display("ko-KR")] Korean,
-            [Display("zh-CN")] ChineseSimplified,
+            [Display("zh-HK")] ChineseTraditional,
```
feliwir: Are you sure that the other one isn't used at all / can be replaced?

src/OpenSage.Game/OpenSage.Game.csproj
```diff
@@ -39,7 +39,7 @@
   <ItemGroup>
     <ShaderFiles Include="Assets\Shaders\*.vert;Assets\Shaders\*.frag" />
     <UpToDateCheckInput Include="@(ShaderFiles)" />
-    
+
```
feliwir: Those newlines probably can be removed again

src/OpenSage.Game/UserSettingsFiles/FontFallbackSettings.json
```diff
@@ -0,0 +1,26 @@
+﻿{
```
feliwir: I'd probably put this file to Content/Translation or Content/Fonts

src/OpenSage.Game/Content/Translation/Providers/CsfTranslationProvider.cs
BSG-75: Even if the other one (simplified Chinese) exist, we don't know it's language code in Generals, so... for now I think it doesn't have a place in this enum, so it should just be replaced

src/OpenSage.Game/UserSettingsFiles/FontFallbackSettings.json
BSG-75: All right, will change it to Content/Fonts

src/OpenSage.Game/OpenSage.Game.csproj
BSG-75: Just checked the project file, it always had new lines, should we remove all extra new lines, or you meant restore those removed spaces so they don't appear in the git diff?

feliwir: Ah you just removed spaces. Didn't look correctly then :) Just keep it as is.

src/OpenSage.Game/Content/Translation/Providers/CsfTranslationProvider.cs
feliwir: Alright

src/OpenSage.Game/OpenSage.Game.csproj
BSG-75: So I'll just keep the "remove space" change?

feliwir: Yes

PR #407, opened 2020-03-17
Fix maps directory not being created for map cache

alandoherty:
`GenerateMapCacheIniFile` assumes the `Maps/` folder exists in the user data directory, added a call to create the directory if it does not exist preventing crash on startup when game not ran before.


Skleni:
Alternatively the call could be skipped when there's no Maps directory...no need to generate an empty file.

feliwir:
> Alternatively the call could be skipped when there's no Maps directory...no need to generate an empty file.

True, but i think the original game always creates one or am i mistaken?

alandoherty:
Added exist check, it seems cleaner to always have the file even if no maps are cached. Saves checks and issues from code expecting it to at least exist even if it is empty.

PR #413, opened 2020-04-14
Feature/textbox

DaanMeijer:
Has a very very dirty implementation of some textbox functionality


src/OpenSage.Game/Gui/Wnd/Images/Image.cs
```diff
@@ -29,6 +29,10 @@ internal void SetSize(in Size size)
             }
 
             _texture = _source.GetTexture(size);
+            if(size.Width == 0 || size.Height == 0)
```
feliwir, line 32: I think this only happens when a Layout call was missing

PR #416, opened 2020-04-22
work in progress on lan lobby joining

feliwir:


src/OpenSage.Game/Network/SkirmishManager.cs
```diff
@@ -0,0 +1,181 @@
...
+using System.Linq;
+using ProtoBuf;
+
+namespace OpenSage.Network
+{
+    public class SkirmishManager
+    {
+        public class SkirmishClient
+        {
+            TcpClient Client;
```
feliwir, line 17: Why are there two clients? This is unused

DaanMeijer: At the moment, it hasn't even been run once, on my machine that is :D So, it's well possible that there is extraneous code :)

PR #417, opened 2020-05-06
Apt embedding

feliwir:
- add more instructions
- allow embedding apts into each other
- some fixes

### DIFF
--- a/src/OpenSage.Game/Gui/Apt/ActionScript/Opcodes/Arithmetic.cs
+++ b/src/OpenSage.Game/Gui/Apt/ActionScript/Opcodes/Arithmetic.cs
@@ -24,7 +24,10 @@ namespace OpenSage.Gui.Apt.ActionScript.Opcodes
 
         public override void Execute(ActionContext context)
         {
-            throw new NotImplementedException();
+            var a = context.Stack.Pop();
+            var b = context.Stack.Pop();
+
+            context.Stack.Push(Value.FromFloat(b.ToFloat() - a.ToFloat()));
         }
     }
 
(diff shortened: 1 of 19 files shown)


src/OpenSage.Game/Gui/Apt/ActionScript/Opcodes/Arithmetic.cs
```diff
@@ -24,7 +24,10 @@ public sealed class Subtract : InstructionBase
 
         public override void Execute(ActionContext context)
         {
-            throw new NotImplementedException();
+            var a = context.Stack.Pop();
+            var b = context.Stack.Pop();
+
+            context.Stack.Push(Value.FromFloat(b.ToFloat() - a.ToFloat()));
```
tgjones, line 30: Is all apt arithmetic in floating point?

src/OpenSage.Game/Gui/Apt/ActionScript/Opcodes/Function.cs
```diff
@@ -326,6 +329,10 @@ public override void Execute(ActionContext context)
             var args = FunctionCommon.GetArgumentsFromStack(context);
 
             FunctionCommon.ExecuteFunction(funcName, args, obj, context);
+
+            var result = context.Stack.Pop();
```
tgjones, line 333: Do you need to check if the function returns a result? Or do all functions return a result?

src/OpenSage.Game/Gui/Apt/ActionScript/Opcodes/Arithmetic.cs
feliwir: Yes. NonNumerical values default to 0 (which ToFloat *should* do, not sure here tbh). See https://www.adobe.com/content/dam/acom/en/devnet/pdf/swf-file-format-spec.pdf **ActionSubtract**

src/OpenSage.Game/Gui/Apt/ActionScript/Opcodes/Function.cs
feliwir: That is actually a very good question. In case there is no return an "Undefined" should be pushed on the stack, which i don't do yet. Let me add this to this PR later.

PR #423, opened 2020-06-11
Minor fixes

feliwir:
- Open archives in read only mode to improve performance
- Fix a test failure when executed multiple times

### DIFF
--- a/src/OpenSage.Game.Tests/OpenSage.Game.Tests.csproj
+++ b/src/OpenSage.Game.Tests/OpenSage.Game.Tests.csproj
@@ -39,7 +39,7 @@
   
   <ItemGroup>
     <None Update="Data\Big\Assets\test.big">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
   </ItemGroup>
 </Project>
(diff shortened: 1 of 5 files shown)


src/OpenSage.Game/Data/FileSystem.cs
```diff
@@ -53,7 +53,7 @@ public FileSystem(string rootDirectory, FileSystem nextFileSystem = null)
 
         private void AddBigArchive(string path)
         {
-            var archive = new BigArchive(path);
+            var archive = new BigArchive(path, BigArchiveMode.Read);
```
tgjones, line 56: Actually I'd prefer the default is changed to `.Read`, not `.Update`. `ZipArchive` defaults to `ZipArchiveMode.Read` for example.

src/OpenSage.Game.Tests/OpenSage.Game.Tests.csproj
```diff
@@ -39,7 +39,7 @@
   
   <ItemGroup>
     <None Update="Data\Big\Assets\test.big">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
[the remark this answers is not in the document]
feliwir: Yeah, multiple runs would fail because the big gets modified. I'll do a copy as you suggested

src/OpenSage.Game/Data/FileSystem.cs
feliwir: True, not sure why i picked Update as the default mode, i was actually looking at the ZipArchive

PR #427, opened 2020-07-15
Add new behavior modules

feliwir:


src/OpenSage.Game/Logic/Object/Upgrade/MaxHealthUpgrade.cs
```diff
@@ -1,7 +1,41 @@
...
+        {
+            if (triggered)
+            {
+                var maxHealthUpgrade = _moduleData as MaxHealthUpgradeModuleData;
+                _gameObject.Body.MaxHealth += (Fix64) maxHealthUpgrade.AddMaxHealth;
+
+                switch (maxHealthUpgrade.ChangeType)
+                {
+                    case MaxHealthChangeType.PreserveRatio:
+                        Fix64 ratio = _gameObject.Body.Health / _gameObject.Body.MaxHealth;
```
tgjones, line 25: I'd have thought this should be `Fix64 ratio = _gameObject.Body.Health / <MaxHealth before it was changed just above>;`

feliwir: Good catch! I'll put the maxHealth assignment below the switch statement

feliwir: Fixed this

PR #428, opened 2020-07-16
Dep updates

feliwir:
- Update Veldrid & ImageSharp

### DIFF
--- a/src/OpenSage.Game/Terrain/Terrain.cs
+++ b/src/OpenSage.Game/Terrain/Terrain.cs
@@ -17,6 +17,7 @@ using OpenSage.Utilities.Extensions;
 using SixLabors.ImageSharp;
 using SixLabors.ImageSharp.PixelFormats;
 using SixLabors.ImageSharp.Processing;
+using SixLabors.ImageSharp.Processing.Processors.Transforms;
 using Veldrid;
 using Veldrid.ImageSharp;
 using Rectangle = OpenSage.Mathematics.Rectangle;
@@ -397,7 +398,7 @@ namespace OpenSage.Terrain
                 {
                     if (tgaFile.Header.Width != largestTextureSize)
                     {
-                        tgaImage.Mutate(x => x.Resize((int) largestTextureSize, (int) largestTextureSize));
+                        tgaImage.Mutate(x => x.Resize((int) largestTextureSize, (int) largestTextureSize, LanczosResampler.Lanczos3));
                     }
 
                     var imageSharpTexture = new ImageSharpTexture(tgaImage);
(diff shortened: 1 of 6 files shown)


src/OpenSage.Game/Terrain/Terrain.cs
```diff
@@ -402,8 +397,7 @@ private struct BlendData
                 {
                     if (tgaFile.Header.Width != largestTextureSize)
                     {
-                        tgaImage.Mutate(x => x
-                            .Resize((int) largestTextureSize, (int) largestTextureSize, MapTextureResampler));
+                        tgaImage.Mutate(x => x.Resize((int) largestTextureSize, (int) largestTextureSize));
```
tgjones, line 400: What happened to the overload that takes an IResampler? I needed to use that because the default one resulted in low quality resampling.

feliwir: For me it looked the same with the default sampler, maybe there was a bug in a previous version?

tgjones: The problem was only visible on terrain textures that were smaller than the standard size. I can’t remember if they were only used on certain maps. I need to check.

feliwir: I fixed it now. The resamplers are static members in ImageSharp itself now

src/OpenSage.Game/Terrain/Terrain.cs
```diff
@@ -402,8 +398,7 @@ private struct BlendData
                 {
                     if (tgaFile.Header.Width != largestTextureSize)
                     {
-                        tgaImage.Mutate(x => x
-                            .Resize((int) largestTextureSize, (int) largestTextureSize, MapTextureResampler));
+                        tgaImage.Mutate(x => x.Resize((int) largestTextureSize, (int) largestTextureSize, LanczosResampler.Lanczos3));
```
tgjones, line 401: Did you intentionally change this from `Lanzos2` to `Lanzos3`? (I have no idea of the pros/cons.)

feliwir: I was reading https://en.wikipedia.org/wiki/Lanczos_resampling and there was a quote from Jim Blinn:
```
According to Jim Blinn, the Lanczos kernel (with a = 3) "keeps low frequencies and rejects high frequencies better than any (achievable) filter we've seen so far."[3]
```

And after opening an discussion on the ImageSharp GitHub discussion page i was told that a range of 3 used to be the default in earlier versions. But we can stick with 2 if you prefer that

tgjones: Nope, that's fine - just wanted to check it was intentional. If Jim Blinn says something's good then we should do that :)

PR #432, opened 2020-07-19
Fix division by zero and index out of bounds

feliwir:
- Fixed in BFME some particlesystems that couldn't use ushort as index type 
- In BFME `PhysicsBehaviors` have no mass. Looking at the ini`s the GravityMult property seems similar


tgjones:
Replaced by #435.

PR #434, opened 2020-07-19
Some tweaks

tgjones:


src/OpenSage.Game/Logic/Object/Upgrade/UpgradeModule.cs
```diff
@@ -96,8 +96,8 @@ public abstract class UpgradeModuleData : BehaviorModuleData
             { "ActiveDuringConstruction", (parser, x) => x.ActiveDuringConstruction = parser.ParseBoolean() },
         };
 
-        public string[] TriggeredBy { get; private set; }
-        public string[] ConflictsWith { get; private set; }
+        public LazyAssetReference<UpgradeTemplate>[] TriggeredBy { get; private set; }
```
tgjones, line 99: @feliwir this is probably useful for you to know about - I added this `LazyAssetReference<T>` thing so that we can create weak references when parsing, which are resolved later when everything is loaded. I'm converting them from `string` fields on-demand when I need them.

feliwir: Yeah, that's cool and seems more elegant

PR #438, opened 2020-07-23
End caps for roads

Skleni:


Skleni:
Courtesy of @bumchen :)

tgjones:
Thank you @bumchen :)

PR #439, opened 2020-07-24
Bfme fixes

feliwir:
- Set initial health correctly for BFME gameobjects
- Prevent projectilebehavior for non-projectiles
- Read more fields of the NamedCamera

### DIFF
--- a/src/OpenSage.Game/Data/Map/NamedCamera.cs
+++ b/src/OpenSage.Game/Data/Map/NamedCamera.cs
@@ -13,25 +13,17 @@ namespace OpenSage.Data.Map
 
         public Vector3 EulerAngles { get; private set; }
         public float Zoom { get; private set; }
+        public Vector2 Unknown { get; private set; }
 
         internal static NamedCamera Parse(BinaryReader reader)
         {
-            var lookAt = reader.ReadVector3();
-            var name = reader.ReadUInt16PrefixedAsciiString();
-            //var rotmat = reader.ReadMatrix3x2();
-
-            var eulerAngles = reader.ReadVector3();
-
-            var zoom = reader.ReadSingle();
-            var unknown2 = reader.ReadSingle();
-            var unknown3 = reader.ReadSingle();
-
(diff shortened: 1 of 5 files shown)


src/OpenSage.Game/Data/Map/NamedCamera.cs
```diff
@@ -9,28 +11,37 @@ public sealed class NamedCamera
         public Vector3 LookAtPoint { get; private set; }
         public string Name { get; private set; }
 
-        public Vector3 Unknown1 { get; private set; }
-        public Vector3 Unknown2 { get; private set; }
+        public Vector3 EulerAngles { get; private set; }
+        public float Zoom { get; private set; }
```
tgjones, line 16: Please could you add `Vector2 Unknown` here? The general pattern is to store the values, so they can be written back out. Otherwise the roundtrip tests will fail.

src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -133,6 +133,7 @@ public string Name
         public Team Team { get; set; }
 
         public bool IsSelectable { get; private set; }
+        public bool isProjectile { get; private set; } = false;
```
tgjones, line 136:
```suggestion
        public bool IsProjectile { get; private set; } = false;
```

src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -250,6 +251,11 @@ public bool IsDamaged
                 _rallyPointMarker = AddDisposable(new GameObject(rpMarkerDef, gameContext, owner, parent));
             }
 
+            if(Definition.KindOf?.Get(ObjectKinds.Projectile) ?? true)
```
tgjones, line 254:
```suggestion
            if (Definition.KindOf?.Get(ObjectKinds.Projectile) ?? false)
```

I think we should default to `false`, shouldn't we?

PR #442, opened 2020-07-26
Avoid reducing mapped images quality

feliwir:
A user on our discord server recognized that Wnd images are a bit blurry / not the full resolution.
There are multiple recognized:
- The texture size depends on the size of the Wnd control, but we always want the full resolution
- The images do get rendered into a framebuffer and then rendered again. I'm not sure if this lowers quality, but it definetly is less efficient than a plain `commandList.CopyTexture` 

I solved this by using the original texture data when possible. However not sure how to solve this for stretchable images. 
Also i think we need to reconsider the **ImageSource** interface. Opinions welcome


tgjones:
Thanks!

PR #443, opened 2020-07-27
Fixed multi selection

Tarcontar:
At the moment units and buildings can be selected at the same time. 
After this fix units are prioritized in order to still be able to set waypoints for them etc.
If multiple structures are in the selected region, only the first one in the list will be selected.

### DIFF
--- a/src/OpenSage.Game/Logic/SelectionSystem.cs
+++ b/src/OpenSage.Game/Logic/SelectionSystem.cs
@@ -158,6 +158,8 @@ namespace OpenSage.Logic
             var boxFrustum = GetSelectionFrustum(SelectionRect);
             var selectedObjects = new List<GameObject>();
 
+            GameObject structure = null;
+
             // TODO: Optimize with frustum culling?
             foreach (var gameObject in Game.Scene3D.GameObjects.Items)
             {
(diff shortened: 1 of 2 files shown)


src/OpenSage.Game/Logic/SelectionSystem.cs
```diff
@@ -173,20 +174,36 @@ private void MultiSelect()
...
                 {
-                    var objectId = (uint) Game.Scene3D.GameObjects.GetObjectId(gameObject);
-                    selectedObjectIds.Add(objectId);
+                    selectedObjects.Add(gameObject);
                 }
             }
 
+
+            var units = selectedObjects.Where(x => x.Definition.KindOf?.Get(ObjectKinds.Structure) == false).ToList();
+            var structures = selectedObjects.Where(x => x.Definition.KindOf?.Get(ObjectKinds.Structure) == true).ToList();
```
tgjones, line 183: You could do `Take(1)` here, since you only ever want the first one.

At some point we'll want to remove LINQ from game-loop code, but this is fine for now.

Tarcontar: I refactored the whole code, should be ready now.

PR #453, opened 2020-08-02
Castle walls fix

Tarcontar:
The walls of fortress maps where all positioned in the center of the fortress, since the obejct property "GeometryRotationAnchorOffset" was not handled properly.
Also fixed the rotation of fortresses (bfme2 starting buildings) placed at the player start positions.

### DIFF
--- a/src/OpenSage.Game/Logic/Object/GameObject.cs
+++ b/src/OpenSage.Game/Logic/Object/GameObject.cs
@@ -23,12 +23,12 @@ namespace OpenSage.Logic.Object
         internal static GameObject FromMapObject(
             MapObject mapObject,
             AssetStore assetStore,
-            GameObjectCollection parent,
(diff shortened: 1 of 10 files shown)


src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -23,12 +23,12 @@ public sealed class GameObject : DisposableBase
         internal static GameObject FromMapObject(
             MapObject mapObject,
             AssetStore assetStore,
-            GameObjectCollection parent,
-            in Vector3 position,
+            GameObjectCollection gameObjects,
```
feliwir, line 26: Why did you rename that?

Tarcontar: since 'parent' was very misleading in that context

PR #457, opened 2020-08-07
Add initial radar implementation

tgjones:
* [x] Draw minimap
* [x] Draw camera frustum
* [x] Handle left / right mouse clicks
* [x] Parse radar events from .sav files


tgjones:
I'll parsed radar events from .sav files, but I'll leave actually implementing them for the future.

feliwir:
Looks good to me apart from the failing tests. I'll add this to Apt, after you've fixed the failing tests

PR #459, opened 2020-08-10
Implement custom cursors

tgjones:
This PR is just the first step - to just render custom cursors.

The next step would be to change the cursor as appropriate for the context of the game.


src/OpenSage.Game/Utilities/Sdl2Interop.cs
```diff
@@ -0,0 +1,130 @@
+﻿using System;
```
feliwir, line 1: I think in the long-run we should try to get some of this into Veldrid.SDL, since people request similar improvements there

PR #480, opened 2020-08-24
Replace SpendMoney() with ReceiveMoney() for selling and cancelling

Evulant:


Tarcontar (approved):
Looking good, thanks!!

PR #482, opened 2020-08-28
Road improvements

### LINKED ISSUE #468: Glitchy stretched models in El Scorcho
Pictures are self explanatory. Could be a terrain rendering issue or other models we're rendering wrong (or shouldn't render to begin with). The map is _El Scorcho_ from Zero Hour.

OpenSage Launcher_2020-08-15_22-40-08
OpenSage Launcher_2020-08-15_22-39-37

Skleni:
Fixes #468

image

I also added a workaround for loading user map textures and implemented a "restart" function (when a map name is passed via the command line, pressing `Break` reloads that map).


tgjones (approved):
Looks good! I don't pretend to understand all of the road curve stuff, which makes me all the more glad that you're working on it 😄

PR #494, opened 2020-09-26
Creation, formation and movement for hordes

Tarcontar:
HordeObjects now create their respective payload objects with the formation as defined in the rankInfo blocks.
Hordes can now be selected by selecting a single payload object.
Hordes move to the target point and rotate accordingly.
Hordes can be produced in buildings (doors open and close accordingly).


Tarcontar:
I will add some more stuff, do not merge yet.

tgjones (approved):

PR #495, opened 2020-09-26
Implement Submixing & 3D Audio

feliwir:
- [X] Update to latest SharpAudio version, which supports 3D Audio & submixers
- [x] Implement submixing in OpenSage
- [x] Implement 3D audio in OpenSage


feliwir:
What needs to be done:
- Track game object position when moving when playing a sound
- Implement proper volume falloff when moving away
- Fix the listener math

tgjones (approved):
Cool, didn’t expect we’d already have some 3D audio support :)

PR #499, opened 2020-10-04
Music playback via Skirmish Scripts

Skleni:
Music playback is controlled via the SkirmishScripts (at least in Generals and Zero Hour). I added code for loading and "instatiating" these scripts (copying them for each player and making sure script-, variable- and team-names are unique) and implemented the script actions and conditions necessary for music playback.

There's probably still a lot missing (e.g. I didn't import teams for now, I don't check if there are already scripts in the map file that need to be preserved, there's no fade-in or fade-out, and so on). But I think the PR is big enough already and basic playback works for Generals and Zero Hour. :)

### DIFF
--- a/src/OpenSage.Game/Audio/AudioSystem.cs
+++ b/src/OpenSage.Game/Audio/AudioSystem.cs
@@ -23,6 +23,9 @@ namespace OpenSage.Audio
 
         private readonly Dictionary<string, int> _musicTrackFinishedCounts = new Dictionary<string, int>();
 
+        private string _currentTrackName;
+        private SoundStream _currentTrack;
+
         public AudioSystem(Game game) : base(game)
         {
             _engine = AudioEngine.CreateDefault();
@@ -48,6 +51,15 @@ namespace OpenSage.Audio
             if (camera != null)
             {
                 UpdateListener(camera);
+            }
+
+            if (_currentTrack != null && !_currentTrack.IsPlaying)
+            {
+                _musicTrackFinishedCounts.TryGetValue(_currentTrackName, out var count);
+                _musicTrackFinishedCounts[_currentTrackName] = count + 1;
+                _currentTrack.Dispose();
+                _currentTrack = null;
+                _currentTrackName = null;
             }
         }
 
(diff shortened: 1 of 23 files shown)


src/OpenSage.Game/Audio/AudioSystem.cs
```diff
@@ -139,7 +160,7 @@ public void PlayAudioEvent(string eventName)
             }
 
             PlayAudioEvent(audioEvent);
-        }
+        }
```
feliwir, line 163: nitpick here: whitespaces should probably be removed

src/OpenSage.Game/Audio/AudioSystem.cs
```diff
@@ -201,5 +222,26 @@ public void PlayAudioEvent(BaseAudioEventInfo baseAudioEvent)
...
+            // TODO: fading
+
+            if (_currentTrack != null)
+            {
+                _currentTrack.Stop();
+                _currentTrack.Dispose();
+            }
+
+            _currentTrackName = musicTrack.Name;
+            _currentTrack = GetStream(musicTrack.File.Value.Entry);
```
feliwir, line 237: I think **GetStream** should also support submixers now. See **GetSound**

tgjones (approved):
This looks great, thank you!

src/OpenSage.Game/Audio/AudioSystem.cs
Skleni: Does `SoundStream` support mixers yet? If yes, how? :)

Skleni: We'll implement this once it's possible.

src/OpenSage.Game/Scripting/ScriptList.cs
```diff
@@ -132,5 +133,15 @@ internal void Load(BinaryReader reader)
...
             }
         }
+
+        internal ScriptList Copy(string appendix)
+        {
+            return new ScriptList()
+            {
+                Scripts = Scripts.Select(s => s.Copy(appendix)).ToArray(),
+                ScriptGroups = ScriptGroups.Select(s => s.Copy(appendix)).ToArray()
+                // TODO: how to set Version correctly?
```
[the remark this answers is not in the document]
Skleni: It doesn't seem to be an issue as long as the scripts aren't written to a file. Maybe this could happen when we write .sav files later on though.

PR #501, opened 2020-10-12
Spawn update, SubObjectUpgrade, LumberMill ressource gathering

Tarcontar:
*  [x] fix quadtree "findNearby"
* [x] spawn initial units based on SpawnUpdate
* [x] show and hide SubObjects based on SubObjectUpgrade
* [x] scan for nearby units in CastleBehavior and unpack respective building
* [x] fix lumbermill worker ressource gathering

### DIFF
--- a/src/OpenSage.Game/Diagnostics/MainView.cs
+++ b/src/OpenSage.Game/Diagnostics/MainView.cs
@@ -138,8 +138,10 @@ namespace OpenSage.Diagnostics
                             {
                                 var random = new Random();
                                 var playableSides = _context.Game.GetPlayableSides().ToList();
-                                var faction1 = playableSides[random.Next(0, playableSides.Count())];
-                                var faction2 = playableSides[random.Next(0, playableSides.Count())];
+                                //var faction1 = playableSides[random.Next(0, playableSides.Count())];
+                                //var faction2 = playableSides[random.Next(0, playableSides.Count())];
+                                var faction1 = playableSides[1];
+                                var faction2 = playableSides[1];
 
                                 if (mapCache.Key.IsMultiplayer)
                                 {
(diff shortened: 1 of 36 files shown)


src/OpenSage.Game/Diagnostics/MainView.cs
```diff
@@ -138,8 +138,10 @@ public void Draw(ref bool isGameViewFocused)
                             {
                                 var random = new Random();
                                 var playableSides = _context.Game.GetPlayableSides().ToList();
-                                var faction1 = playableSides[random.Next(0, playableSides.Count())];
-                                var faction2 = playableSides[random.Next(0, playableSides.Count())];
+                                //var faction1 = playableSides[random.Next(0, playableSides.Count())];
+                                //var faction2 = playableSides[random.Next(0, playableSides.Count())];
+                                var faction1 = playableSides[3];
```
feliwir, line 143: Why is that change?

src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -545,12 +561,7 @@ public bool UpgradeAvailable(UpgradeTemplate upgrade)
 
         public bool ConflictingUpgradeAvailable(UpgradeTemplate upgrade)
         {
-            if (upgrade == null)
-            {
-                return false;
-            }
-
-            if (upgrade.Type == UpgradeType.Player)
+            if (upgrade == null || upgrade.Type == UpgradeType.Player)
```
feliwir, line 564: When UpgradeType is Player, it should probably check the owner for conflicting upgrades

src/OpenSage.Game/Diagnostics/MainView.cs
Tarcontar: because for testing it is a pain in the ass when the faction is randomly chosen :)

src/OpenSage.Game/Logic/Object/GameObject.cs
Tarcontar: There is already a todo for that in the line below, just combined both checks into a single line here

feliwir (approved):

PR #504, opened 2020-10-21
Bfme upgrades

Tarcontar:
- cancel unit build command
- attribute modifiers (production, health)
- PassiveAreaEffectBehavior
- fixed rohan peasant horde
- reduced gameObject collider updates and Quadtree access
- conflicting player upgrades
- hide build foundation plot as long as a structure is build on it
- CommandSetUpgrade
- initial player upgrades from playertemplate.ini
- check for collissions only on moved objects
- gettingBuildCommandset for buildings under construction

### DIFF
--- a/src/OpenSage.Game/Content/SubsystemLoader.cs
+++ b/src/OpenSage.Game/Content/SubsystemLoader.cs
@@ -190,6 +190,7 @@ namespace OpenSage.Content
                         case SageGame.Bfme:
                         case SageGame.Bfme2:
                         case SageGame.Bfme2Rotwk:
+                            _contentManager.LoadIniFile(@"Data\INI\Attributemodifier.ini");
                             _contentManager.LoadIniFile(@"Data\INI\Experiencelevels.ini");
                             break;
                     }
--- a/src/OpenSage.Game/Content/SubsystemLoader.cs
+++ b/src/OpenSage.Game/Content/SubsystemLoader.cs
@@ -108,6 +108,9 @@ namespace OpenSage.Content
                 case Subsystem.InGameUI:
                     LoadFiles(@"Data\INI\InGameUI.ini");
                     break;
+                case Subsystem.AttributeModifiers:
+                case Subsystem.ExperienceLevels:
+                    break;
                 default:
                     throw new ArgumentOutOfRangeException(nameof(subsystem), subsystem, null);
             }
--- a/src/OpenSage.Game/Content/SubsystemLoader.cs
+++ b/src/OpenSage.Game/Content/SubsystemLoader.cs
@@ -184,18 +184,6 @@ namespace OpenSage.Content
                     }
                     break;
 
-                case Subsystem.ObjectCreation:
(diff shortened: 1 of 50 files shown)


src/OpenSage.Game/Content/SubsystemLoader.cs
```diff
@@ -108,6 +108,9 @@ public void Load(Subsystem subsystem)
                 case Subsystem.InGameUI:
                     LoadFiles(@"Data\INI\InGameUI.ini");
                     break;
+                case Subsystem.AttributeModifiers:
```
feliwir, line 111: I think those subsystems should not be loaded in the first place for Generals. See ContentManager.cs

src/OpenSage.Game/Diagnostics/MainView.cs
```diff
@@ -140,8 +140,8 @@ public void Draw(ref bool isGameViewFocused)
                                 var playableSides = _context.Game.GetPlayableSides().ToList();
                                 //var faction1 = playableSides[random.Next(0, playableSides.Count())];
                                 //var faction2 = playableSides[random.Next(0, playableSides.Count())];
-                                var faction1 = playableSides[3];
-                                var faction2 = playableSides[3];
+                                var faction1 = playableSides[0];
```
feliwir, line 143: This still seems super weird. A faction picker would make more sense

src/OpenSage.Game/Logic/Object/Behaviors/SlowDeathBehavior.cs
```diff
@@ -29,7 +29,7 @@ internal SlowDeathBehavior(SlowDeathBehaviorModuleData moduleData)
             _moduleData = moduleData;
         }
 
-        internal bool IsApplicable(DeathType deathType) => _moduleData.DeathTypes?.Get(deathType) ?? false;
+        internal bool IsApplicable(DeathType deathType) => _moduleData.DeathTypes?.Get(deathType) ?? true;
```
feliwir, line 32: Why you changed this to true? I think the default must be **false**

Tarcontar: Because some models, like e.g. the hero statue of Rohan in bfme only have a single slowDeathBehavior which does not define any death types.

src/OpenSage.Game/Diagnostics/MainView.cs
Tarcontar: Go for it:)

src/OpenSage.Game/Content/SubsystemLoader.cs
Tarcontar: done

src/OpenSage.Game/Diagnostics/MainView.cs
Tarcontar: done

feliwir (approved):

PR #505, opened 2020-10-22
Bfme apt improvements

Tarcontar:


src/OpenSage.Game/Gui/Apt/ActionScript/Opcodes/Basic.cs
```diff
@@ -101,7 +100,9 @@ public sealed class Delete : InstructionBase
 
         public override void Execute(ActionContext context)
         {
-            throw new NotImplementedException();
+            var value = context.Pop();
```
feliwir, line 103: This is not correct according to SWF specification:
https://www.adobe.com/content/dam/acom/en/devnet/pdf/swf-file-format-spec.pdf

I think you should first **Pop** the name, then the object and then deletect the variable from that object

feliwir (changes requested):

src/OpenSage.Game/Gui/Apt/ActionScript/Opcodes/Basic.cs
Tarcontar: i removed it for now

feliwir (approved):

PR #509, opened 2020-11-15
Migrate to net 5

feliwir:


.github/workflows/ci.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     - name: Install .NET Core
```
Tarcontar, line 30: should we remove 'Core' here?

Tarcontar (approved):

PR #513, opened 2020-12-21
Process lua code in AnimationStates

Tarcontar:
- parse lua code in AnimationStates
- hide drawModules accordingly

### DIFF
--- a/src/OpenSage.Game/Logic/Object/GameObject.cs
+++ b/src/OpenSage.Game/Logic/Object/GameObject.cs
@@ -216,7 +216,26 @@ namespace OpenSage.Logic.Object
 
         public int Supply { get; set; }
 
-        public List<string> HiddenSubObjects;
+        private List<string> _hiddenSubObjects;
+        private List<string> _shownSubObjects;
+
+        public void HideSubObject(string subObject)
+        {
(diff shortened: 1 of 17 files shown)


src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -216,7 +216,41 @@ public void DoDamage(DamageType damageType, Fix64 amount, DeathType deathType, T
...
 
-        public List<string> HiddenSubObjects;
+        private Dictionary<string, bool> _hiddenSubObjects;
+        private Dictionary<string, bool> _shownSubObjects;
+
+        public void HideSubObject(string subObject, bool permanent = false)
+        {
+            if (!_hiddenSubObjects.ContainsKey(subObject))
+            {
+                _hiddenSubObjects.Add(subObject, permanent);
```
tgjones, line 226: With this logic, if there’s an existing object, it won’t have it’s “permanent” value updated with the new value passed in here. Is that intended?

tgjones (approved):

src/OpenSage.Game/Logic/Object/GameObject.cs
Tarcontar: indeed, should be fixed now

PR #514, opened 2020-12-21
work on intersection stuff

Tarcontar:
- lots of work on collision (including tests)
- support for looping and stopping AudioEvents
- do not render objects outside of camera frustum
- GateOpenAndCloseBehavior for Bfme1 fortress gates
- show colliders in debug overlay
- BuildingBehavior
- CastleBehavior -> check for nearby structure and hide corresponding object
- dedicated TransitionStates for CncGenerals and Bfme1 (and ongoing) since they used to inherit from ModelConditionState and now inherit from AnimationState
- GeometryUpgrade
- update Colliders for buildPreviewObject
- improved object selection (using all colliders of an object)
- improved TerrainPassabilityUpdates

### DIFF
--- a/src/OpenSage.Game/Audio/AudioSystem.cs
+++ b/src/OpenSage.Game/Audio/AudioSystem.cs
@@ -161,6 +161,20 @@ namespace OpenSage.Audio
             PlayAudioEvent(audioEvent);
         }
 
+        public void DisposeSource(AudioSource source)
+        {
+            if (source == null)
+            {
+                return;
+            }
+
+            if (source.IsPlaying())
+            {
+                source.Stop();
+            }
+            _sources.Remove(source);
+        }
+
         private bool ValidateAudioEvent(BaseAudioEventInfo baseAudioEvent)
         {
             if (baseAudioEvent == null)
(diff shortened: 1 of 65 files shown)


src/OpenSage.Game/Audio/AudioSystem.cs
```diff
@@ -161,6 +161,20 @@ public void PlayAudioEvent(string eventName)
...
+            if (source == null)
+            {
+                return;
+            }
+
+            if (source.IsPlaying())
+            {
+                source.Stop();
+            }
+            _sources.Remove(source);
```
feliwir, line 175: This could crash if the AudioSource was not created by the AudioSystem

feliwir (commented):
Looks good to me apart from one note. I think @tgjones  or @paavohuhtala  should check this aswell

src/OpenSage.Game/Audio/AudioSystem.cs
Tarcontar: No, `Remove` would only return `False` if the source was not in the list.

feliwir (approved):

PR #516, opened 2020-12-28
FIx audio crashes

feliwir:


src/OpenSage.FileFormats.Big/BigArchiveEntryStream.cs
```diff
@@ -17,9 +16,6 @@ public BigArchiveEntryStream(BigArchiveEntry entry, uint offset)
             _entry = entry;
             _archive = entry.Archive;
             _offset = offset;
-
-            _archive.AcquireLock();
```
feliwir, line 21: Not sure why the archive got locked when opening a stream

src/OpenSage.FileFormats.Big/BigArchiveEntryStream.cs
```diff
@@ -33,6 +29,7 @@ public override void Flush()
 
         public override int Read(byte[] buffer, int offset, int count)
         {
+            _archive.AcquireLock();
```
feliwir, line 32: Only locking during a read now

Tarcontar (approved):

PR #517, opened 2021-01-02
Updated SharpAudio Version

voldien:
Updates the SharpAudio version dependency that will allow systems that use OpenAL as their Audio SubSystem, such as Linux, to start. Previously it wouldn't start at all.


feliwir (approved):
Thanks for your contribution (especially the on in SharpAudio!)

PR #518, opened 2021-01-04
Disable RenderDoc menu item if renderdoc failed to load.

voldien:
If the renderdoc failed to load, the program will crash if pressing on any of the renderdoc menu items. By disabling the renderdoc menu item it will prevent the program to crash but still show that the option was enabled.


feliwir (approved):
Looks good to me! Thanks a lot

PR #519, opened 2021-01-04
Online multiplayer

Skleni:
Implemented an online multiplayer mode by reusing the existing Direct Connect window. So for now, there is no lobby, but you can host a game and join it by entering the host's IP address.
In order to support clients behind NAT, the Open.Nat package was added to automatically forward the necessary ports via UPnP (if supported by the router).
The existing network code was refactored and improved (e.g. changing the map is now supported) .


src/OpenSage.Game/Game.cs
```diff
@@ -746,9 +780,10 @@ public void HostSkirmishGame()
                 new EchoConnection(),
                 null,
                 0,
-                isMultiPlayer: false);
+                isMultiPlayer: false,
+                seed: Environment.TickCount);
         }
-
+
```
feliwir, line 786: nitpick. Please remove

feliwir (commented):
Looks good to me. I think @tgjones or someone else should have a look at this aswell

src/OpenSage.Game/Data/Sav/SaveFile.cs
```diff
@@ -110,8 +110,9 @@ public static void LoadFromStream(Stream stream, Game game)
                                             game.AssetStore.PlayerTemplates.GetByName("FactionAmerica"), // TODO
                                             new ColorRgb(0, 0, 255))
                                     },
-                                    0, // TODO
-                                    false,
+                                    localPlayerIndex: 0, // TODO
+                                    isMultiPlayer: false,
+                                    seed: Environment.TickCount, // TODO
```
Skleni, line 115: I added a parameter for the random seed to the `StartMultiPlayerGame` method (as all players need to use the same seed). This caused quite a lot of changes in various files. `Environment.TickCount` is used as a default in non-multiplayer scenarios.

Skleni:
There is still lots of room for improvement and I want to implement some more stuff, but it works (now also with more than 2 players :) ) and the PR is big enough already.

src/OpenSage.Core/NLog.config
```diff
@@ -15,6 +15,7 @@
 
   <rules>
     <logger name="OpenSage.Scripting.*" writeTo="" final="true"/>
+    <logger name="OpenSage.Network.*" minlevel="Trace" writeTo="console" final="true"/>
```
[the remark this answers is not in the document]
feliwir: Seems fine for now to me

src/OpenSage.Mods.Generals/Gui/MainMenuCallbacks.cs
```diff
@@ -140,6 +142,13 @@ void OpenSinglePlayerSideMenu(string side, string sideWindowSuffix)
                             context.WindowManager.SetWindow(@"Menus\LanLobbyMenu.wnd");
                             break;
 
+                        case "MainMenu.wnd:ButtonOnline":
+                            // this should load another window, but we don't have a lobby server yet,
+                            // so we just show the "Direct Connect" window
+
+                            context.WindowManager.SetWindow(@"Menus\NetworkDirectConnect.wnd", NetworkUtils.OnlineTag);
```
feliwir, line 149: We need to keep in mind that this. Needs to be changed. Maybe add a "TODO:" , so it's easier to find later on?

feliwir (approved):

PR #520, opened 2021-01-08
Add Name for Sampler Objects

voldien:
Add name for the sampler objects to make it easier to debug in OpenGL at least.


Tarcontar (approved):

Tarcontar:
Thx!

PR #521, opened 2021-01-08
Fix issue with weird UI Rendering on OpenGL

voldien:
The OpenGL rendering of the UI didn't always rendering probably such as the main menu. It seems to to have been the cause by the sprite fragment shader. It would call textureSize(AlphaMask_1, 0) for getting the size for computing the UV for the getting the alphaMaskColor. The AlphaMask_1 sampler2D is just a dummy created by the glslangValidator and is never bounded texture object.

### DIFF
--- a/src/OpenSage.Game/Assets/Shaders/Sprite.frag
+++ b/src/OpenSage.Game/Assets/Shaders/Sprite.frag
@@ -1,7 +1,5 @@
 #version 450
 
-#extension GL_EXT_samplerless_texture_functions : enable
-
 #define FILL_METHOD_NORMAL     0
 #define FILL_METHOD_RADIAL_360 1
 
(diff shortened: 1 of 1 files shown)


src/OpenSage.Game/Assets/Shaders/Sprite.frag
```diff
@@ -74,11 +74,11 @@ void main()
 
     // Currently we always sample from the alpha mask,
     // falling back to a default one if there isn't a real one.
-    ivec2 alphaMaskSize = textureSize(AlphaMask, 0);
+    ivec2 alphaMaskSize = textureSize(sampler2D(AlphaMask, Sampler), 0);
```
Tarcontar, line 77: I seems like the `GL_EXT_samplerless_texture_functions` - extension does not work for OpenGL then?

voldien: I am not sure, I just checked my RX 580 and gtx1070 and neither of them has that extension, so perhaps that explains why glslangValidator created a dummy texture object AlphaMask_1 that was never bound.

Tarcontar: ok, i think we should remove that extension at the top then, since we do not use it anymore in that file.

Tarcontar (approved):

PR #526, opened 2021-01-18
Trying to handle loading of custom maps

BSG-75:
Let OpenSage search for maps in `%appdata%\{UserDataLeafName}\Maps` for games from CNC3 onward.

### DIFF
--- a/src/OpenSage.Game/Content/ContentManager.cs
+++ b/src/OpenSage.Game/Content/ContentManager.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.IO;
 using System.Text;
 using OpenSage.Content.Translation;
 using OpenSage.Data;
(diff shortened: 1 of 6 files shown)


src/OpenSage.Game/Content/ContentManager.cs
```diff
@@ -157,15 +160,22 @@ internal void LoadIniFile(FileSystemEntry entry)
         internal FileSystemEntry GetMapEntry(string mapPath)
         {
             var normalizedPath = FileSystem.NormalizeFilePath(mapPath);
-            if (UserDataFileSystem != null && normalizedPath.StartsWith(UserDataFileSystem.RootDirectory))
+
+            if (SageGame >= SageGame.Cnc3)
```
feliwir, line 164: Wasn't the check in UserMapsFilesystem meant to prevent that check here?

BSG-75: Yes, but since we use virtual paths for games after CNC3, the ` mapPath.Substring(UserDataFileSystem.RootDirectory.Length + 1)` is not needed anymore.
But probably it doesn't harm either, I'll try to remove this `if` check, and see how it runs.

Tarcontar (approved):

PR #530, opened 2021-02-26
Add teams and configure allies/enemies

charliefoxtwo:
This utilized the teams configured in a skirmish lobby and populates each player's allies/enemies.

### DIFF
--- a/src/OpenSage.Game/Logic/PlayerSettings.cs
+++ b/src/OpenSage.Game/Logic/PlayerSettings.cs
@@ -21,9 +21,9 @@ namespace OpenSage.Logic
         public PlayerTemplate Template { get; set; }
         public PlayerOwner Owner { get; set; }
         public string Name { get; set; }
-        public byte Team { get; set; }
+        public int Team { get; set; }
 
-        public PlayerSetting(int? startPosition, PlayerTemplate template, ColorRgb color, byte team, PlayerOwner owner = PlayerOwner.None, string name = "")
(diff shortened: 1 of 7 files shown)


charliefoxtwo:
I started working on showing the proper cursors, but realized I couldn't tell if an object was friendly or not so I switched over to get this one done. I've tested these changes with that I have so far for the cursor implementation and they seem to work well enough.

src/OpenSage.Game/Logic/PlayerSettings.cs
```diff
@@ -21,14 +21,16 @@ public struct PlayerSetting
         public PlayerTemplate Template { get; set; }
         public PlayerOwner Owner { get; set; }
         public string Name { get; set; }
+        public byte Team { get; set; }
 
-        public PlayerSetting(int? startPosition, PlayerTemplate template, ColorRgb color, PlayerOwner owner = PlayerOwner.None, string name = "")
+        public PlayerSetting(int? startPosition, PlayerTemplate template, ColorRgb color, byte team, PlayerOwner owner = PlayerOwner.None, string name = "")
```
feliwir, line 26: We usually don't use **byte** as a numeric type. Just use **int** here, since it makes no big difference and avoids any future limitations

charliefoxtwo: Sure thing. I swear there was a reason I used byte (I thought something else was already configured with it?) but skimming the code now I can't figure out what it was. Anyway, I agree int is the reasonable solution here.

charliefoxtwo:
Feel free to squash this when merged.

Tarcontar (approved):

Tarcontar:
thx!

PR #533, opened 2021-02-27
Add cursor logic

charliefoxtwo:
Looking for comments and input on this. I'm sure I've missed some areas (or maybe even gotten some parts wrong) so I'd appreciate a few sets of eyes on this one.

### DIFF
--- a/src/OpenSage.Mods.Generals/Logic/OrderGenerators/GeneralsOrderGenerator.cs
+++ b/src/OpenSage.Mods.Generals/Logic/OrderGenerators/GeneralsOrderGenerator.cs
@@ -1,0 +1,513 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Numerics;
+using FixedMath.NET;
+using OpenSage.Graphics.Cameras;
+using OpenSage.Graphics.Rendering;
+using OpenSage.Input;
+using OpenSage.Logic;
+using OpenSage.Logic.Object;
+using OpenSage.Logic.OrderGenerators;
+using OpenSage.Logic.Orders;
+
+namespace OpenSage.Mods.Generals.Logic.OrderGenerators;
+
+internal sealed class GeneralsOrderGenerator : IOrderGenerator
+{
+    private readonly Game _game;
+
+    private Vector3 _worldPosition;
+    private GameObject _worldObject;
+
+    private IReadOnlyCollection<GameObject> SelectedUnits => LocalPlayer.SelectedUnits;
+    private Player LocalPlayer => _game.PlayerManager.LocalPlayer;
+
+    public bool CanDrag { get; } = false;
+
+    public GeneralsOrderGenerator(Game game)
+    {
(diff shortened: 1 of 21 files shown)


charliefoxtwo:
Added some of the logic implementation. I think it's worth considering having the `GetCursor()` method return something that indicates the order type that should be processed if the player were to click, as the alternative is basically to run very similar logic again when the player clicks to determine what order to give. If this is being run every tick it should always be correct, and storing something like an enum should be a minimal cost compared to the savings of not having to duplicate the logic and track changes in both places.

tgjones (approved):
This looks really great, thank you! And the documentation is really appreciated too.

charliefoxtwo:
The one concern I found with this as I was working on OpenSAGE more is that this is a rather complex logic tree that would have to be largely duplicated when we determine what action a unit should take if the user clicks where the cursor is. I don't think it's a reason not to move forward with this, but just something to think about.

charliefoxtwo:
Updated this so it should at least build and run now

charliefoxtwo:
Chinooks currently get the enter vehicle cursor when hovering over other chinooks, so some work is definitely still required lol

charliefoxtwo:
Dozers also get an attack cursor when hovering over supply warehouses -_- I suspect this is partly due to them being considered enemy buildings, but either way a dozer shouldn't be able to attack anything.

charliefoxtwo:
Updated to move order generator logic to game-specific projects. While the games may share some cursors there are also plenty more they don't share, and this should be encapsulated on a per-game basis. Given that the logic for showing the `GenericInvalid` cursor at times seems to be different between even Generals and Zero Hour (see documentation for more details), I _suspect_ this is the right move.

We can (and likely should) still move some of the newly-created methods (see `GeneralsOrderGenerator.cs`) to base classes so that other games' respective order generators can utilize this behavior.

charliefoxtwo:
Two years later, I'm finally ready to merge this 🙈 

It will likely conflict with https://github.com/OpenSAGE/OpenSAGE/pull/695 but fixing the conflict shouldn't be too bad.

src/OpenSage.Mods.Generals/Logic/OrderGenerators/GeneralsOrderGenerator.cs
```diff
@@ -0,0 +1,492 @@
...
+        // target is some game object
+        if (TargetIsEnemy(target))
+        {
+            return GetCursorForEnemyTarget(target);
+        }
+
+        // target is not enemy
+        if (AnySelectedUnitCanCaptureTarget(target))
+        {
+            return GeneralsCursors.CaptureBuilding;
```
charliefoxtwo, line 119: I was just digging into the ini files some more and noticed that for `CommandButton Command_AmericaRangerCaptureBuilding` there is a specified `CursorName` as well as `InvalidCursorName`. Special powers, fire weapon commands, etc should probably use whatever is defined there instead of being hardcoded here.

charliefoxtwo:
Superseded by #855

PR #535, opened 2021-03-02
Fix #532 add null check on CurrentWeapon in TurretAIUpdate

### LINKED ISSUE #532: Building GLA Radar Van crashes the game
https://github.com/OpenSAGE/OpenSAGE/blob/396f907d7c59f78b6dc5ee646316b9a8903da6be/src/OpenSage.Game/Logic/Object/Update/AIUpdate/TurretAIUpdate.cs#L42-L49

The cause appears to be an NPE on line 42 (and a subsequent one on 48) where `CurrentWeapon` is null (presumably since the Radar Van doesn't have any weapons). I've changed it locally (just to `CurrentWeapon?.CurrentTarget`) and it seems fine, but I haven't looked too closely to see if this has other potential ramifications, so I'd appreciate some input on this one.

On the topic of NPEs, has any thought been given to enabling C#'s nullable reference types? I know it's a bit divisive in the community, but at least personally I really like the language feature.

charliefoxtwo:
Vehicles like the GLA Radar Van have no Weapon Set defined, and thus CurrentWeapon will never be anything but null for them.

### DIFF
--- a/src/OpenSage.Game/Logic/Object/Update/AIUpdate/TurretAIUpdate.cs
+++ b/src/OpenSage.Game/Logic/Object/Update/AIUpdate/TurretAIUpdate.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using OpenSage.Data.Ini;
 using OpenSage.Mathematics;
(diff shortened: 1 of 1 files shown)


Tarcontar (approved):

Tarcontar:
thx!

PR #536, opened 2021-03-03
Fix right click camera pan behavior

charliefoxtwo:
This should bring the right-click camera pan system more closely in line with Generals/ZH.

### DIFF
--- a/src/OpenSage.Game/Logic/SelectionSystem.cs
+++ b/src/OpenSage.Game/Logic/SelectionSystem.cs
@@ -38,7 +38,7 @@ namespace OpenSage.Logic
         private Point2D _selectionEndPoint;
 
         private Point2D _panStartPoint;
-        private readonly Stopwatch _panStopwatch = new Stopwatch();
+        private TimeSpan _panStartTime;
 
         public SelectionStatus Status { get; private set; } = SelectionStatus.NotSelecting;
         public bool Selecting => Status != SelectionStatus.NotSelecting;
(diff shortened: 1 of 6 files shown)


Tarcontar:
This is looking good for me. I have no idea how this is affected by the multiple? control options in generals (we recently switched to the old version).

charliefoxtwo:
Are you referring to the alternate click mode (right click to issue move orders, etc)? I was thinking about that one yesterday... Will require more thought to implement though. Or are you referring to something else?

src/OpenSage.Game/Logic/SelectionSystem.cs
```diff
@@ -23,20 +24,32 @@ public enum SelectionStatus
...
+
         private SelectionGui SelectionGui => Game.Scene3D.SelectionGui;
 
-        private Point2D _startPoint;
-        private Point2D _endPoint;
+        private Point2D _selectionStartPoint;
+        private Point2D _selectionEndPoint;
+
+        private Point2D _panStartPoint;
+        private readonly Stopwatch _panStopwatch = new Stopwatch();
```
charliefoxtwo, line 41: Looking through more of the solution, I'm actually not convinced using a stopwatch is the best course of action here. Should I be using `TimeInterval` and `Game.MapTime` instead? That seems to be the established pattern elsewhere.

Tarcontar: Indeed stopwatch is not the preferred way to do it.

Tarcontar:
> Are you referring to the alternate click mode (right click to issue move orders, etc)? I was thinking about that one yesterday... Will require more thought to implement though. Or are you referring to something else?

Yes that was what i was referring to, i am not an expert tho. @tgjones might know more about that.

src/OpenSage.Game/Logic/SelectionSystem.cs
feliwir: Yes, we don't use StopWatches anywhere usually

charliefoxtwo: I'll get that changed, thanks

charliefoxtwo:
CI build failure seems to be unrelated?

Let me know if the updated timing code is any better. I mimicked what I saw in `GameObject.Construct()`, so hopefully that's the right idea.

feliwir:
> CI build failure seems to be unrelated?

CI failure was unrelated, sorry about that

> Let me know if the updated timing code is any better. I mimicked what I saw in `GameObject.Construct()`, so hopefully that's the right idea.

Yep, that looks good now. I think we can merge this if @Tarcontar  approves

Tarcontar (approved):

Tarcontar:
Looks good to me, thx!

PR #538, opened 2021-03-11
Add AnimationInstance unit tests

charliefoxtwo:
A good portion of these tests don't pass currently. I've been playing with the animation code a little bit and these should be fairly painless fixes, but it would be good to get more eyes on these tests first to verify they're correct.
image

### DIFF
--- a/src/OpenSage.Game.Tests/Graphics/Animation/AnimationInstanceTests.cs
+++ b/src/OpenSage.Game.Tests/Graphics/Animation/AnimationInstanceTests.cs
@@ -1,0 +1,225 @@
+﻿using System;
+using System.Numerics;
+using OpenSage.Graphics;
+using OpenSage.Graphics.Animation;
+using OpenSage.Logic.Object;
+using Xunit;
+
+namespace OpenSage.Tests.Graphics.Animation
+{
+    public class AnimationInstanceTests
+    {
+        [Fact]
+        public void OnceForwardsTest()
+        {
+            var modelBoneInstances = NewDefaultModelBoneInstances();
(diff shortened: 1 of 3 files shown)


charliefoxtwo:
All tests should be passing now, but @Tarcontar if you could take a look at 44cdccc it would be much appreciated as I basically just deleted some stuff to get the tests passing and that may not be the right thing to do.

I'd also appreciate anybody's eyes on the changes in fb00446 as I'm just not certain if that's how we should be solving this problem.

src/OpenSage.Mods.Generals/Gui/GameOptionsUtil.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
```
charliefoxtwo, line 1: This file has changes because I was working off of a commit prior to the latest on master but which had some other changes that have already been merged to master. Should be inconsequential but I can cherry-pick the commits over to a new branch if necessary.

charliefoxtwo:
@Tarcontar iirc last time we looked at this BFME relied on the `Manual` logic - would you be able to test to see if it still works with the changes I've proposed here (basically, also checking for `AdjustHeightByConstructionPercent`).

charliefoxtwo:
Existing behavior:
https://gyazo.com/8461d5c3f7bcede4096c266842f194e5

Fixed behavior in this PR:
https://gyazo.com/5a1e65df95e3122ea1dc9fc513d32115

I tried really hard to embed gifs in this comment but gave up 🤷

feliwir (approved):

PR #545, opened 2021-04-03
Skirmish refactoring

Skleni:
Skirmish games are now handled in the same way as LAN games. Previously, they worked differently with the state (player positions, colors, teams, ...) being only implicitly stored in the UI controls. Now they use the same model (`SkirmishGameSettings`, formerly known as `SkirmishGame`) as LAN games, which makes the code cleaner and simpler.

Based on that, the starting position setup now also works for network games.

### DIFF
--- a/src/OpenSage.Game/Gui/Wnd/Controls/ComboBox.cs
+++ b/src/OpenSage.Game/Gui/Wnd/Controls/ComboBox.cs
@@ -258,7 +258,7 @@ namespace OpenSage.Gui.Wnd.Controls
                 0,
                 ClientSize.Height,
                 ClientSize.Width,
-                itemsHeight);
+                itemsHeight + 2);
         }
 
         protected override void Dispose(bool disposeManagedResources)
--- a/src/OpenSage.Game/Network/SkirmishGameSettings.cs
+++ b/src/OpenSage.Game/Network/SkirmishGameSettings.cs
@@ -2,18 +2,20 @@
 
 namespace OpenSage.Network
 {
-    public class SkirmishGame
+    public class SkirmishGameSettings
     {
+        public const int MaxNumberOfPlayers = 8;
+
         private static readonly NLog.Logger Logger = NLog.LogManager.GetCurrentClassLogger();
 
         private bool _isDirty;
         private string _mapName;
         private SkirmishGameStatus _status;
 
-        public SkirmishGame(bool isHost)
+        public SkirmishGameSettings(bool isHost)
         {
             IsHost = isHost;
-            Slots = Enumerable.Range(0, 8).Select(i => new SkirmishSlot(i)).ToArray();
+            Slots = Enumerable.Range(0, MaxNumberOfPlayers).Select(i => new SkirmishSlot(i)).ToArray();
             Status = SkirmishGameStatus.Configuring;
         }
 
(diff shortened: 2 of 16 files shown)


src/OpenSage.Game/Network/SkirmishManager.cs
```diff
@@ -13,20 +13,61 @@ namespace OpenSage.Network
...
+
+        public abstract bool IsStartButtonEnabled();
+        public abstract Task HandleStartButtonClickAsync();
+
+        public virtual void Stop()
+        {
+        }
+    }
+
+    public sealed class LocalSkirmishManager : SkirmishManager
```
Skleni, line 40: I added a `LocalSkirmishManager` class for non-network skirmish games (next to the existing implementations for Host and Client).

src/OpenSage.Game/Network/SkirmishManager.cs
```diff
@@ -13,20 +13,61 @@ namespace OpenSage.Network
...
+
+        public override Task HandleStartButtonClickAsync()
+        {
+            Connection = new EchoConnection();
+            Settings.Status = SkirmishGameStatus.ReadyToStart;
+            return Task.CompletedTask;
+        }
+    }
+
+    public abstract class NetworkSkirmishManager : SkirmishManager
```
Skleni, line 59: This is now the base class for the Host and Client skirmish managers (the new `LocalSkirmishManager` doesn't need the network stuff).

src/OpenSage.Mods.Generals/Gui/GameOptionsUtil.cs
```diff
@@ -16,33 +18,29 @@ namespace OpenSage.Mods.Generals.Gui
 {
     public class GameOptionsUtil
```
Skleni, line 19: This class now contains most of the logic shared between `SkirmishGameOptions` and `LanGameOptions`. It was there before, but was quite a mess, with half the logic existing there and the other half being in the UI Callbacks.

src/OpenSage.Game/Network/SkirmishGameSettings.cs
```diff
@@ -2,18 +2,20 @@
 
 namespace OpenSage.Network
 {
-    public class SkirmishGame
+    public class SkirmishGameSettings
```
Skleni, line 5: I figured `SkirmishGameSettings` was a better name (`SkirmishGame` being rather ambiguos).

src/OpenSage.Game/Gui/Wnd/Controls/ComboBox.cs
```diff
@@ -258,7 +258,7 @@ protected override void LayoutOverride()
                 0,
                 ClientSize.Height,
                 ClientSize.Width,
-                itemsHeight);
+                itemsHeight + 2);
```
feliwir, line 261: Why was this required?

src/OpenSage.Game/Network/SkirmishGameSettings.cs
feliwir: Agreed :)

feliwir (approved):

src/OpenSage.Game/Gui/Wnd/Controls/ComboBox.cs
Skleni: Not entirely sure...I added code to not draw ListBoxItems that are outside of the visible area of the ListBox. This lead to the last item not being drawn without these 2 added pixels...I guess there's some kind of border or padding involved.

PR #554, opened 2021-04-26
Add GameLogic and GameClient

tgjones:
This PR adds `GameLogic` and `GameClient`, and moves drawing-related code from `GameObject` into a new `Drawable` class. Eventually we'll move more from `Scene3D` into `GameLogic` and `GameClient`.

The names `GameLogic` and `GameClient` come from the `CHUNK_GameLogic` and `CHUNK_GameClient` chunks in .sav files. The `Drawable` name comes from many mentions to "objects and drawables", where objects contain logic and drawables contain drawing stuff, in .ini files. Also in .sav files, objects contain behaviours, while drawables contain draw modules and client updates.

### DIFF
--- a/src/OpenSage.Game/Client/Drawable.cs
+++ b/src/OpenSage.Game/Client/Drawable.cs
@@ -1,0 +1,291 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using System.Numerics;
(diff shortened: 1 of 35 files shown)


feliwir (approved):
Looks great, thanks

PR #557, opened 2021-05-01
Fix MapCache load error on Linux

### LINKED ISSUE #543: OS dependant path separator for key name breaks on Linux
Paths in `.big` files uses backslash.

This means during `.big` parsing, that `MapCache` entry have a name with `maps\shellmap1\shellmap.map` (and a `instanceId` with a hash generated by this name).

`ResolveToVirtualPath()` use `NormalizeFilePath()`. `NormalizeFilePath()` replace separator by OS dependant separator, this means on Linux, `Maps\ShellMap1\ShellMap1.map` become `maps/shellmap1/shellmap.map`.

This means when `GetByName()` is called and generate a hash for `maps/shellmap1/shellmap.map`, it can't find `MapCache` entry for `maps\shellmap1\shellmap.map` in `_assetScopes[0].Asset`.

wrightjjw:
Fixes #543. `GetByName()` expects a Windows filepath, so it changes `mapPath` to one before calling it.

### DIFF
--- a/src/OpenSage.Game/OpenSage.Game.csproj
+++ b/src/OpenSage.Game/OpenSage.Game.csproj
@@ -60,6 +60,18 @@
     <GlslangValidatorPath Condition="$([MSBuild]::IsOsPlatform('linux'))">linux-x64\glslangValidator</GlslangValidatorPath>
   </PropertyGroup>
 
+<PropertyGroup Condition="'$(GlslangValidatorPath)'=='win-x64\glslangValidator.exe'">
+  <DefineConstants>_WINDOWS_</DefineConstants>
+</PropertyGroup>
+
+<PropertyGroup Condition="'$(GlslangValidatorPath)'=='osx-x64\glslangValidator'">
+  <DefineConstants>_MACOSX_</DefineConstants>
+</PropertyGroup>
+ 
+<PropertyGroup Condition="'$(GlslangValidatorPath)'=='linux-x64\glslangValidator'">
+  <DefineConstants>_LINUX_</DefineConstants>
+</PropertyGroup>
+
   <ItemGroup>
     <None Update="Content\Fonts\FontFallbackSettings.json" CopyToOutputDirectory="PreserveNewest" />
   </ItemGroup>
--- a/src/OpenSage.Game/Scene3D.cs
+++ b/src/OpenSage.Game/Scene3D.cs
@@ -308,10 +308,9 @@ namespace OpenSage
             if (mapPath != null)
             {
                 // Change filepath separator from / to \ when not on windows (GetByName() expects windows path)
-                if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-                {
+                #if !_WINDOWS_
                     mapPath = mapPath.Replace("/", "\\");
-                }
+                #endif
                 var mapCache = game.AssetStore.MapCaches.GetByName(mapPath.ToLower());
                 if (mapCache == null)
                 {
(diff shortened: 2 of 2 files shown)


andreinonea:
This check happens at run-time. I noticed in `OpenSage.Game/OpenSage.Game.csproj`, there are already platform checks, albeit for another reason.

```
<PropertyGroup>                                                               
  <GlslangValidatorPath>win-x64\glslangValidator.exe</GlslangValidatorPath>   
  <GlslangValidatorPath Condition="$([MSBuild]::IsOsPlatform('osx'))">osx-x64\glslangValidator</GlslangValidatorPath>
  <GlslangValidatorPath Condition="$([MSBuild]::IsOsPlatform('linux'))">linux-x64\glslangValidator</GlslangValidatorPath>
</PropertyGroup>
```                                                          
Thus, I propose a simple addition to the `.csproj` file that creates defines for each platform.

```
<PropertyGroup Condition="'$(GlslangValidatorPath)'=='win-x64\glslangValidator.exe'">
  <DefineConstants>_WINDOWS_</DefineConstants>                                
</PropertyGroup>                                                              
 
<PropertyGroup Condition="'$(GlslangValidatorPath)'!='win-x64\glslangValidator.exe'">
  <DefineConstants>_UNIX_</DefineConstants>                                   
</PropertyGroup>                                                              
 
<PropertyGroup Condition="'$(GlslangValidatorPath)'=='osx-x64\glslangValidator'">
  <DefineConstants>_MACOSX_</DefineConstants>                                 
</PropertyGroup>                                                              
 
<PropertyGroup Condition="'$(GlslangValidatorPath)'=='linux-x64\glslangValidator'">
  <DefineConstants>_LINUX_</DefineConstants>                                  
</PropertyGroup>
```
Then, we could simply write

```
if (mapPath != null)                                                
{                                                                   
#if _UNIX_                                                          
    mapPath = mapPath.Replace("/", "\\");                           
#endif                                                              
    var mapCache = game.AssetStore.MapCaches.GetByName(mapPath.ToLower());
    ...
    ...
```
in `OpenSage.Game/Scene3D.cs`, which is even more readable and also has the advantage of happening at compile-time.

Please note that I **have NOT tested** this on my machine because of another bug which prevents me from starting the Launcher, so I might be wrong. It all goes under the assumption that `$([MSBuild]::IsOsPlatform())` does work indeed in this scenario.

wrightjjw:
That's definitely a better idea. I'll test it out when I get a chance.

wrightjjw:
Made changes based on @andreinonea's suggestion. Similar result to original solution.

dginovker:
Ty for fixing up Linux :heart_eyes:. It wasn't even starting for me a few months ago <3

tgjones (changes requested):
Hello! I'm sorry for not reviewing this PR before. I _think_ it would be better to instead change this line:

https://github.com/OpenSAGE/OpenSAGE/blob/c0c8940c7c181745c27bd1acb64ccad4253e147d/src/OpenSage.Game/Content/MapCache.cs#L12

to this:

``` csharp
(x, name) => x.SetNameAndInstanceId("MapCache", FileSystem.NormalizeFilePath(IniParser.ToAsciiString(name))),
```

In general, all paths used at runtime are supposed to use the correct path separators for the current platform. But that is one place where we missed it, because the `MapCache` keys themselves are paths.

wrightjjw:
Thanks for looking into it. Yeah, that fix is a lot simpler and better. I'm quite busy for a couple days but I will look into this sometime in the next week.

wrightjjw:
Opening a clean branch to fix this.

cbordeman:
You don't need a Windows define for this, the .Net APIs and nearly all Win32 APIs can use forward slash in paths without issue.  They have since XP at least.

PR #559, opened 2021-05-02
german installations do not have a lang folder

Tarcontar:


feliwir:
As DarthJane suggested we could also use the SplashScreen for BFME games for language detection if that is more consistent

andreinonea:
Neither strategies produce fail-safe code. There are three issues here.

1. Splash files should not be used because they may be missing without breaking the game.
2. <_lang_>\_Audio.big files should not be used because in Bfme2 they do not exist for several languages, namely Dutch, Norwegian, Polish and Swedish.
3. This code in `DetectFromFileSystem` will always fail for all Bfme games when the installed language is not Dutch, because EnglishAudio is present no matter what language you choose to install.

```
            var files = Directory.GetFiles(rootDirectory, $"{filePrefix}*{fileSuffix}", SearchOption.TopDirectoryOnly) // there's no sense in searching subfolders
                .Select(x => Path.GetFileName(x))
                .Select(x => string.IsNullOrEmpty(filePrefix) ? x : x[filePrefix.Length..])
                .Select(x => string.IsNullOrEmpty(fileSuffix) ? x : x[..^fileSuffix.Length]);
            foreach (var file in files)
            {
                if (file.Length == 0)
                {
                    continue;
                }
                return file;
            }
```

The for loop returns as soon as one match is found. As D < E, it will properly find Dutch, but all other languages start with a letter >= E so it always return English.

The only consistent file for language detection is <_lang_>\.big. I will try to think of a better solution, but one would be to have an array of languages (there are 9 in total for Bfme games) and use something like this:

```
        private static readonly string[] GameLanguages = new string[]
        {
            "dutch",
            "english",
            "french",
            "german",
            "italian",
            "norwegian",
            "polish",
            "spanish",
            "swedish"
        };

        private static string DetectLanguage(string langDirectory)
        {
            foreach (var lang in GameLanguages)
            {
                if (File.Exists(Path.Combine(langDirectory, lang + ".big")))
                {
                    return lang;
                }
            }
            return DefaultLanguage;
        }
        ..
        ..
// ReadCurrentLanguage
                case SageGame.Bfme:
                    return DetectLanguage(rootDirectory);
                case SageGame.Bfme2:
                case SageGame.Bfme2Rotwk:
                    return DetectLanguage(Path.Combine(rootDirectory, "lang"));
```

This has been tested and produces expected results for every existing language.
Can someone with access to Command & Conquer games check that the same language file rules apply there?

lanyizi:
@andreinonea newer C&C games uses Windows Registry and Skudef files to decide which files (including localization files) to load. So you could have multiple language pack installed, the game will only load the needed files for the current language.

I'm not sure about Generals / BFME because I don't play them too much, but apparently the issue we're talking here is Generals/BFME specific?

PR #578, opened 2021-05-30
Refactor FileSystem

tgjones:
The main problem I wanted to solve here was efficiently loading files from a specific directory, which is done at startup for loading object ini files. Our previous implementation was not directory-centric at all, so loading files from a specific directory meant doing a `StartsWith` on every single entry's path.

This implementation fixes that, and also separates more cleanly the implementations for the local file system and the .big file system.


feliwir:
The tests don't compile currently if you missed it

Skleni:
While you're working on the `FileSystem`: I had to add the following call in order to load user maps:
image

tgjones:
> I had to add the following call in order to load user maps

@Skleni do you mean loading user maps via `--map`? Because we haven't implemented the "Unofficial Maps" button yet, right?

Skleni:
> > I had to add the following call in order to load user maps
> 
> @Skleni do you mean loading user maps via `--map`? Because we haven't implemented the "Unofficial Maps" button yet, right?

Yes, exactly. I think I also added the user maps to the jump list some time ago, but it doesn't seem to work atm.

src/OpenSage.Game/Content/Translation/TranslationManager.cs
```diff
@@ -212,59 +214,54 @@ public string GetParticularString(string context, string str)
...
+                case SageGame.Cnc3:
+                case SageGame.Cnc3KanesWrath:
+                    path = "cnc3";
+                    break;
+                case SageGame.Ra3:
+                case SageGame.Ra3Uprising: // there is a data/gamestrings_temp.csf in Uprising
+                case SageGame.Cnc4:
+                    path = Path.Combine("data", "gamestrings");
+                    break;
+            }
```
Qibbi, line 246: May I suggest moving that into the game's definition?

tgjones: Good idea - [done](https://github.com/OpenSAGE/OpenSAGE/pull/578/commits/60b22023b18eb999952b5bbde356c88f6250d5ee).

tgjones:
Going to do this a different way.

PR #586, opened 2021-06-06
Implement Parse and WriteTo for SkyboxSettings

DarkAtra:
The test map was created using Worldbuilder for BFME 1 and is empty, but has the following SkyboxSettings:
grafik


feliwir (approved):
Thanks a lot for your work! Looks great

PR #587, opened 2021-06-09
Readd coverage calculation

feliwir:
Eventhough our testing isn't fully functional yet in CI (only a small amount of tests executed), it has become super easy with .NET 5 to calculate coverage statistics.

### DIFF
--- a/src/OpenSage.Game.CodeGen/ScriptActionsGenerator.cs
+++ b/src/OpenSage.Game.CodeGen/ScriptActionsGenerator.cs
@@ -33,7 +33,7 @@ namespace OpenSage.Game.CodeGen
             sb.AppendLine("    }");
             sb.AppendLine("}");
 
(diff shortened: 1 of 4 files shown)


src/OpenSage.Game.CodeGen/ScriptActionsGenerator.cs
```diff
@@ -33,7 +33,7 @@ public override void Execute(GeneratorExecutionContext context)
             sb.AppendLine("    }");
             sb.AppendLine("}");
 
-            context.AddSource("ScriptActions.Execution.generated.cs", sb.ToString());
+            context.AddSource("ScriptActions.Execution.g.cs", sb.ToString());
```
feliwir, line 36: Required since this seems to be the default extension that **Coverlet** recognizes. It then knows that this is generated by  source generator. See https://github.com/coverlet-coverage/coverlet/issues/1084#issuecomment-857549034

tgjones (approved):

PR #588, opened 2021-06-09
Use fixedmath nuget

feliwir:
Use a NuGet package instead of embedding the source. See
https://github.com/OpenSAGE/FixedMath.NET


feliwir:
need to fix the namespaces inside the FixedMath package.... But it actually was like that

tgjones (approved):

PR #591, opened 2021-06-22
Parse ScriptsPlayers properties

Skleni:
Adds conditional parsing of player properties for ScriptsPlayers.

### DIFF
--- a/src/OpenSage.Game/Data/Scb/ScriptPlayer.cs
+++ b/src/OpenSage.Game/Data/Scb/ScriptPlayer.cs
@@ -1,0 +1,30 @@
+﻿using System.IO;
+using OpenSage.Data.Map;
+using OpenSage.FileFormats;
+
+namespace OpenSage.Data.Scb
+{
+    public sealed class ScriptPlayer
+    {
+        public string Name { get; private set; }
+        public AssetPropertyCollection Properties { get; private set; }
+
+        internal static ScriptPlayer Parse(BinaryReader reader, MapParseContext context, bool parseProperties)
+        {
+            return new ScriptPlayer()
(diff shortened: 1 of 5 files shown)


src/OpenSage.Game/Data/Scb/ScriptPlayer.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.IO;
+using OpenSage.Data.Map;
+using OpenSage.FileFormats;
+
+namespace OpenSage.Data.Scb
+{
+    public sealed class ScriptPlayer
```
Skleni, line 7: Or should this be Script*s*Player?

src/OpenSage.Game/Data/Scb/ScriptsPlayers.cs
```diff
@@ -9,64 +8,67 @@ public sealed class ScriptsPlayers : Asset
...
+                ushort unknown = 0;
+                if (version < 2)
                 {
-                    throw new InvalidDataException();
+                    unknown = reader.ReadUInt16();
+                    if (unknown != 0)
+                    {
+                        throw new InvalidDataException();
+                    }
                 }
```
Skleni, line 36: I couldn't find this value in Zero Hour maps/scripts.

src/OpenSage.Game/Data/Scb/ScriptPlayer.cs
tgjones: I guess so, if you don’t mind renaming it, otherwise I’m happy to merge as is.

tgjones (approved):
Thank you!

PR #600, opened 2021-08-15
Fix skirmish menu slots

charliefoxtwo:
The color slot was defaulting to gold and the team slot was defaulting to 1. This also disables slot editing when a "player" (human or AI) is not present.


src/OpenSage.Mods.Generals/Gui/GameOptionsUtil.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
charliefoxtwo, line 1: Not sure what's going on here? Perhaps a BOM added or removed?

Skleni (approved):
Thank you!

PR #603, opened 2021-09-08
Use mmap in readonly mode for big archives

feliwir:
Since the **BIG** format is a memory mapped fileformat we can use memmap functionality in **READ** mode. There isn't a large difference in terms of performance or memory usage, but i guess that's mostly because we/C# performs a copy of the mmap for every **Read**. Maybe someone has advice on how to use mmap efficiently in C#


smx-smx:
You could use C# Spans together with MemoryMappedFile. I implemented it in this project of mine: https://github.com/smx-smx/PDBSharp/blob/master/PDBSharp/MemoryMappedSpan.cs
With this you can create Spans from a memory mapped file without having to create additional copies of the file contents

PR #604, opened 2021-09-15
Add memory mapped spans and streams

smx-smx:
Code is originating from [PDBSharp](https://github.com/smx-smx/PDBSharp/tree/master/PDBSharp) and has undergone some uncommitted/local improvements such as Write mode and Big Endian structure reading while being used in multiple side projects.

Being a separate component, it has been developed under its own code style and license (but i'm open to re-licensing or dual-licensing, as long as i can reuse the shared code).

While it could be refactored to better suit OpenSage, **a possible (and attractive) alternative that i'm considering is to create a Nuget library** that OpenSage (and other projects) can consume

The main feature offered here is the ability to create `Memory<byte>`, and therefore `Span<byte>`, from a `MemoryMappedFile`.
This effectively enables true random access files in C# without having to use intermediate copies or buffered streams.

A subclass of `Stream`, `SpanStream`, makes it possible to use this system as a drop-in replacement in some scenarios (mostly Read-only parsing, as growing of `SpanStream` is not supported).

This PR currently enables this scenario for BIG files

Feel free to leave comments, suggestions or feature requests 🙂

### DIFF
--- a/src/OpenSage.Core/Endianness.cs
+++ b/src/OpenSage.Core/Endianness.cs
@@ -1,0 +1,29 @@
+﻿#region License
+/*
+ * Copyright (C) 2019 Stefano Moioli <smxdev4@gmail.com>
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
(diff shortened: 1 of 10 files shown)


smx-smx:
Nuget package published, and BIG files are implemented (both R/W supported).

PR #625, opened 2022-01-09
.sav parsing: JetAIUpdate and Drawables without GameObjects

Skleni:


src/OpenSage.Game.Tests/Data/Sav/SaveFileTests.cs
```diff
@@ -43,6 +43,20 @@ public void CanLoadSaveFiles(string relativePath)
             game.EndGame();
         }
 
+        [Fact]
+        public void CanLoadTestWeapons()
```
tgjones, line 47: Is it necessary to add this test? I think it should be covered by `CanLoadSaveFiles` above.

tgjones (approved):
This is awesome!

src/OpenSage.Game.Tests/Data/Sav/SaveFileTests.cs
Skleni: Right, I added it because the other ones were disabled, but I guess it can be removed.

tgjones: Ah yes, locally I have this change:

``` csharp
//[Theory(Skip = "Doesn't work yet")]
[Theory]
[MemberData(nameof(GetSaveFiles))]
public void CanLoadSaveFiles(string relativePath)
```

tgjones: I think we can enable those tests by default soon - for me locally, there are only 3 failing ones, which happen to be failing on a road-related issue, not a .sav parsing issue ;-)

PR #627, opened 2022-01-16
Fix skudef extension file not found for RA3 and Uprising

voldien:
Fixes the problem of not finding the skudef file for RA3 and Uprising, by ignoring the letter case, Additionally normalizing the file paths constructed from the skudef file.

Also added to extract the default language for RA3 and Uprising.

### DIFF
--- a/src/OpenSage.Game/Data/SkudefReader.cs
+++ b/src/OpenSage.Game/Data/SkudefReader.cs
@@ -1,4 +1,5 @@
-﻿using System;
+using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 
@@ -51,7 +52,9 @@ namespace OpenSage.Data
 
         public static void Read(string rootDirectory, Action<string> addBigArchive)
         {
-            var skudefFiles = Directory.GetFiles(rootDirectory, "*.skudef");
+
+            var skudefFiles = Directory.GetFiles(rootDirectory).Where(filename => filename.EndsWith("skudef", StringComparison.OrdinalIgnoreCase)).ToList();
+
             var skudefFile = skudefFiles
                 .OrderBy(SkudefVersion.Parse)
                 .LastOrDefault(); // TODO: This is not the right logic. needs to take into account the language.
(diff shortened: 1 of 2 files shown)


src/OpenSage.Game/Data/SkudefReader.cs
```diff
@@ -51,7 +53,9 @@ public int CompareTo(SkudefVersion other)
 
         public static void Read(string rootDirectory, Action<string> addBigArchive)
         {
-            var skudefFiles = Directory.GetFiles(rootDirectory, "*.skudef");
+
+            var skudefFiles = Directory.GetFiles(rootDirectory).Where(filename => filename.EndsWith("skudef", StringComparison.OrdinalIgnoreCase)).ToList();
```
lanyizi, line 57: If I have a file named as `RA3_english_1.12skudef` instead of `RA3_english_1.12.skudef`, I think it will be matched by `filename.EndsWith` too

src/OpenSage.Game/Utilities/LanguageUtility.cs
```diff
@@ -39,6 +39,9 @@ public static string ReadCurrentLanguage(IGameDefinition gameDefinition, FileSys
                 case SageGame.Bfme2:
                 case SageGame.Bfme2Rotwk:
                     return DetectFromFileSystem(fileSystem, "lang", "", "Audio.big");
+                case SageGame.Ra3Uprising:
+                case SageGame.Ra3:
+                    return DetectFromFileSystem(fileSystem, "Data", "", "Audio.big");
```
lanyizi, line 44: Some languages doesn't have their own localized audio files, IIRC Traditional Chinese has `ChineseT.big` in RA3, but it still uses `EnglishAudio.big` for audio data

src/OpenSage.Game/Data/SkudefReader.cs
tgjones: We should use the built-in overload of `Directory.GetFiles` that uses case-insensitive matching:

```suggestion
            var skudefFiles = Directory.GetFiles(rootDirectory, "*.skudef", new EnumerationOptions { MatchCasing = MatchCasing.CaseInsensitive };
```

tgjones (changes requested):

tgjones (approved):

PR #668, opened 2022-07-16
Move W3D rendering to separate csproj

tgjones:


tgjones:
I quite like some of the ideas here, but it's unmergeable as-is, so I'll close this. Maybe I'll try to do some of this in future, smaller, PRs.

PR #685, opened 2022-11-05
Get zero hour running

charliefoxtwo:
Just a couple of NPEs prevented zero hour from running. I figured there's no harm in putting some checks around them and getting it running, even if the controlbar is a mess.


charliefoxtwo:
Not sure why or how this one failed the mac build 🤔

tgjones (approved):

PR #687, opened 2022-11-06
Enable development on ubuntu

charliefoxtwo:
This gets Generals running on ubuntu. The cursors bit was a bit puzzling. There might be a cleaner way to do that, I'm not sure.

Of note, the audio can be choppy at times on my machine and the unit select voice lines are very choppy. Otherwise, everything else seems to work exactly as it does on Windows!

### DIFF
--- a/docs/developer-guide.md
+++ b/docs/developer-guide.md
@@ -18,6 +18,38 @@ dotnet build
 
 You can also build like that on Windows, in case you don't want to use Visual Studio.
 
+#### Linux dependencies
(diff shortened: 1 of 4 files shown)


charliefoxtwo:
mac build timed out after 6 hours. Guessing the mac builds are just broken?

feliwir (approved):

charliefoxtwo:
Is it possible to requeue the Mac CI build?

PR #696, opened 2022-11-15
Show/Hide supplies carried by China supply truck

charliefoxtwo:
Currently the China supply truck's supplies are always visible in the back. This is a combination of a few issues:
 - The model being called `Cargo01` in some places and `CARGO01` in others
 - `HideSubObject` and `ShowSubObject` being considered non-animation properties, while adding scripts that only run if the object is considered as having animation properties
 - `W3dModelDraw.SetActiveAnimationState()` returning early before executing lua scripts which handle show/hide sub-objects

@tgjones this might step on some of the work you did earlier this year and may not fit your long-term plan so if you'd prefer this were accomplished a different way let me know.


feliwir:
Resolve conflicts please

src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
```diff
@@ -120,6 +120,11 @@ protected virtual bool SetActiveAnimationState(AnimationState animationState, Ra
                 return false;
             }
 
+            if (animationState?.Script is not null)
```
feliwir, line 123: i'd prefer `!= null`. I don't think we use the other notation anywhere

charliefoxtwo: It is used elsewhere (11 usages in 7 files), but not as prevalent as `!= null`. The same goes for `is null` vs `== null`. I'm fine using `!= null` if that's the preferred style :)

charliefoxtwo:
It might be nice to figure out what's going on with some of these mac CI builds. I've seen them fail a couple of times now and the issue seems to be transient and inconsistent.

Tarcontar (approved):

Tarcontar:
@charliefoxtwo i am working on the macos build, found one issue already. On my m1 macbook air it works right now, so will need some more time to hunt it down. Here seems to be also an issue with the windows build, can you reproduce it?

charliefoxtwo:
> The active test run was aborted. Reason: Test host process crashed : Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at Veldrid.Sdl2.Sdl2Native.SDL_Init(Veldrid.Sdl2.SDLInitFlags)
   at OpenSage.Platform.Start()
   at OpenSage.Tests.GameFixture..ctor()

I have seen this one occur intermittently in the CI builds before. I suspect if it were to be re-run (maybe a couple of times) it would work.

charliefoxtwo:
Updated the branch and it's passing now.

ghost (approved):

PR #718, opened 2023-05-29
Fix #691: Add scroll cursors

### LINKED ISSUE #691: Generals: Holding right click should show arrows based on cursor offset from center
These are:
 - `sccscroll0.ani`
 - `sccscroll1.ani`
 - `sccscroll2.ani`
 - `sccscroll3.ani`
 - `SCCScroll4.ani`
 - `SCCScroll5.ani`
 - `SCCScroll6.ani`
 - `SCCScroll7.ani`

None of these seem to be present in `AssetStore.MouseCursors`.

krasnykon:
Show scroll cursors when camera is moving by mouse or keyboard.  Also fix some other camera and mouse cursor issues:
1. Reset mouse cursor to arrow, when game is unfocused in developer mode
2. Reduce RMB camera movement speed
3. Allow camera moving by hovering mouse cursor at window border

### DIFF
--- a/src/OpenSage.Game/Graphics/Cameras/RtsCameraController.cs
+++ b/src/OpenSage.Game/Graphics/Cameras/RtsCameraController.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Numerics;
 using ImGuiNET;
 using OpenSage.Data.Sav;
+using OpenSage.Input.Cursors;
 using OpenSage.Mathematics;
 using OpenSage.Terrain;
 using Veldrid;
@@ -33,8 +34,8 @@ namespace OpenSage.Graphics.Cameras
 
         public bool CanPlayerInputChangePitch { get; set; }
 
-        private CameraPanDirection _panDirection = CameraPanDirection.None;
-        CameraPanDirection ICameraController.PanDirection { get => _panDirection; }
+        private CursorDirection? _panDirection = null;
+        CursorDirection? ICameraController.PanDirection { get => _panDirection; }
 
         private float _yaw;
         public void SetLookDirection(Vector3 lookDirection)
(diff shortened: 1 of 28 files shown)


src/OpenSage.Game/Input/Cursors/CursorManager.cs
```diff
@@ -35,7 +35,13 @@ public CursorManager(AssetStore assetStore, ContentManager contentManager, GameW
 
         public void SetCursor(string cursorName, in TimeInterval time)
         {
-            if (!_cachedCursors.TryGetValue(cursorName, out var cursor))
+            SetCursor(cursorName, 0, time);
+        }
+
+        public void SetCursor(string cursorName, uint direction, in TimeInterval time)
```
feliwir, line 41: Could direction stay an enum here? I think the casting to an integer should happen inside this method

src/OpenSage.Game/IPanel.cs
feliwir, line 1: The only implementor seems to be `GamePanel`. Why can we not use GamePanel directly for now?

src/OpenSage.Game/Graphics/Cameras/RtsCameraController.cs
```diff
@@ -160,31 +167,66 @@ private static float GetKeyMovement(in CameraInputState inputState, Key positive
...
+
+                    if (inputState.LastX < PanBorderWith) right = -1;
+                    else if (inputState.LastX > bounds.Width - PanBorderWith) right = 1;
+                    else right = GetKeyMovement(inputState, Key.Right, Key.Left);
+
+                    if (inputState.LastY < PanBorderWith) up = 1;
+                    else if (inputState.LastY > bounds.Height - PanBorderWith) up = -1;
+                    else up = GetKeyMovement(inputState, Key.Up, Key.Down);
+                }
+                if (up != 0 || right != 0)
```
feliwir, line 198: Couldn't you check if the absolute value of abs(up)/abs(down) is greater than the threshold? Floats are likely to not be exactly 0 in many cases

src/OpenSage.Game/Graphics/Cameras/RtsCameraController.cs
```diff
@@ -160,31 +167,66 @@ private static float GetKeyMovement(in CameraInputState inputState, Key positive
...
+                    }
+                    else if (up < -thresh)
+                    {
+                        if (right > thresh) _panDirection = CameraPanDirection.RightDown;
+                        else if (right < -thresh) _panDirection = CameraPanDirection.LeftDown;
+                        else _panDirection = CameraPanDirection.Down;
+                    }
+                    else
+                    {
+                        if (right >= 0) _panDirection = CameraPanDirection.Right;
```
feliwir, line 215: there is no threshold for right & left only for up & down?

src/OpenSage.Game/Gui/Wnd/WndInputMessageHandler.cs
```diff
@@ -76,9 +76,10 @@ bool GetControlAtPoint(in Point2D mousePosition, out Control control, out Point2
                                 new WndWindowMessage(WndWindowMessageType.MouseMove, control, mousePosition),
                                 context);
                         }
-                        return mouseOverControls.Count > 0
+                        /*return mouseOverControls.Count > 0
```
feliwir, line 79: I think that change does cause the mouse to click "through" controls? Why was this necessary?

feliwir (changes requested):
Hey @krasnykon ,
thanks a lot for looking at this problem! I have some minor comments & questions, would appreciate if you have time to answer.
It's been a while since i've looked into most of these code parts, so i might miss some obvious things
Thanks!

src/OpenSage.Game/Input/Cursors/CursorManager.cs
krasnykon: Theoretically, there may be cursors that have other directions layout, like having four directions. So think the casting is necessary here.

src/OpenSage.Game/IPanel.cs
krasnykon: For unit testing mostly. Actually I'm about to write unit test for RtsCameraController and add more interface classes

src/OpenSage.Game/Graphics/Cameras/RtsCameraController.cs
krasnykon: Actually values will be 0, when camera isn't moving by user. But you are right, that's the bad way to detect camera movement.

src/OpenSage.Game/Graphics/Cameras/RtsCameraController.cs
krasnykon: That's original game behavior. Scroll cursor always visible when right mouse pressed (by default is right arrow). So threshold  needs only to distinguish vertical movement from horizontal.

src/OpenSage.Game/Gui/Wnd/WndInputMessageHandler.cs
krasnykon: No, controls I checked works normally. What that change do is prevent a mouse hover stealing by wnd controls. If there is some control that broken by this change, it should be treated specifically.

PR #724, opened 2024-01-08
Update to .NET 8

charliefoxtwo:


src/OpenSage.Launcher/OpenSage.Launcher.csproj
```diff
@@ -2,15 +2,15 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <RuntimeIdentifiers>win-x64;osx-x64;ubuntu-x64</RuntimeIdentifiers>
+    <RuntimeIdentifiers>win-x64;osx-x64;linux-x64</RuntimeIdentifiers>
```
charliefoxtwo, line 5: changed from ubuntu-x64 as .net 8 removes distro-specific runtime identifiers https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/rid-graph

src/OpenSage.Game/Graphics/Cameras/Camera.cs
```diff
@@ -131,8 +131,10 @@ private void UpdateProjection()
 
         private void CreateProjection(out Matrix4x4 projection)
         {
+            var fov = GetVerticalFieldOfView();
+            fov = float.IsNaN(fov) ? 1 : fov; // .NET 8 no longer allows CreatePerspectiveFieldOfView to be called with NaN FOV
```
charliefoxtwo, line 135: I'm guessing this function never should have allowed NaN FOV, but did at one point? This seems to only happen on startup when the window size hasn't been updated and is 0x0. This feels like a sketchy hack to get it working so I'm open to alternatives, or maybe we fix it properly down the line.

charliefoxtwo: I'll investigate, may take some time though.

charliefoxtwo: Alright, I believe I've figured it out. I've rolled back the changes in this file.

src/OpenSage.Game/Game.cs
```diff
@@ -532,7 +532,7 @@ public Game(GameInstallation installation)
                 OrderGenerator = AddDisposable(new OrderGeneratorSystem(this));
 
                 Panel.ClientSizeChanged += OnPanelSizeChanged;
-                OnPanelSizeChanged(this, EventArgs.Empty);
+                Panel.EnsureFrame(window.ClientBounds);
```
charliefoxtwo, line 535: Setting the frame size here ensures we never attempt to construct a camera with 0 height (and thus a NaN FOV)

Issue #730, opened 2024-01-12
Generals: Canceling a USA building spawns a ranger


charliefoxtwo:
Rangers should only be spawned on USA building destruction, and the number varies depending on the structure.

Looks like the Ranger spawn is defined by `<faction>OCL_AmericanRangerDebris<id>`. Guessing this is just being called on cancellation. It seems like it also maybe isn't spawning the correct count, so maybe that parameter is being ignored?

it looks like there's an `ExemptStatus` we should be factoring in
```ini
Behavior             = CreateObjectDie ModuleTag_12
  CreationList  = AirF_OCL_AmericanRangerDebris02
  ExemptStatus  = UNDER_CONSTRUCTION
End
```

PR #733, opened 2024-01-13
Respect RequireStatus and ExemptStatus in die behaviors

charliefoxtwo:
This utilizes the `RequiredStatus` and `ExemptStatus` fields in the die behaviors which have at least partial implementations. It does not implement those fields in other behaviors (die or otherwise) which do not have implementations.

- `IsCorrectStatus` is defined in two separate places in this PR. I opted not to extract it out further as I don't know how the other non-die behaviors will use it (if at all). There are currently two other non-die behaviors that have these object status properties to my knowledge (`FireWeaponWhenDeadBehavior` and `FireWeaponCollide`).
- This adds an `ObjectStatus` field to `GameObject`, but does not persist it in any way. I _assume_ one of the unknown fields is this (how else would you persist whether a vehicle has been hijacked?), but I'm not sure. The other Generals/ZH ObjectStatus values all have equivalents in `ModelConditionFlag` (though `UNDER_CONSTRUCTION` would be computed based on one of a few condition flags) - is it possible we're missing one? I tried to load a save file to inspect it, but it seems save file loading is broken at the moment.

At any rate, this results in no rangers spawned when building construction is canceled.

### DIFF
--- a/src/OpenSage.Game/Logic/Object/BehaviorModule.cs
+++ b/src/OpenSage.Game/Logic/Object/BehaviorModule.cs
@@ -9,7 +9,7 @@ namespace OpenSage.Logic.Object
     {
         internal virtual void Update(BehaviorUpdateContext context) { }
 
-        internal virtual void OnDie(BehaviorUpdateContext context, DeathType deathType) { }
+        internal virtual void OnDie(BehaviorUpdateContext context, DeathType deathType, ObjectStatus? status) { }
 
         internal virtual void OnDamageStateChanged(BehaviorUpdateContext context, BodyDamageType fromDamage, BodyDamageType toDamage) { }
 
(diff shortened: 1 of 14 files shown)


charliefoxtwo:
Resolved the merge conflict on mobile and it doesn't look right. Going to take a look in a bit.

charliefoxtwo:
Conflict properly resolved, and caught and fixed what could have been a minor bug with respect to a ModelConditionState flag not getting updated in all instances.

charliefoxtwo:
Fixes #730

Tarcontar (approved):

charliefoxtwo:
Is this one waiting on anything?

Tarcontar (approved):

PR #736, opened 2024-01-15
Fix can't purchase PLAYER_UPGRADEs

charliefoxtwo:


charliefoxtwo:
> The active test run was aborted. Reason: Test host process crashed : Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.

Assuming this isn't related to my changes?

Tarcontar (approved):

PR #774, opened 2024-01-19
Format hotkeys in strings with bold + yellow

charliefoxtwo:
This feels way too far down to put this formatting (surely there are strings we don't want to format?) but I'm not sure where else to put it. I tried `Control`, but building the texture requires information about how large each individual texture is. I then wanted to do something that maybe took multiple `TextKey`s, but that wasn't really a good match either as caching would become an issue.

The solution here _works_, but it doesn't feel like it's in the right place. Does anybody have some ideas for where this could go, or want to take it on themselves?

This solution also only supports a single format character, which is fine for all the strings I checked in Generals but may not be fine for other games.

image


Tarcontar:
@charliefoxtwo what is the status of this? Could this be used as is right now?

charliefoxtwo:
image
image
image

Looks like there's currently a minor issue where spacing between words is lost when the start of the next word is this special character. I'll take a look at that, because that doesn't seem to be an issue when the special character is in the middle of a word.

charliefoxtwo:
This seems to be due to whitespace being trimmed in the current implicitly-installed version of SixLabors.Fonts (1.0.0-beta15).
https://github.com/SixLabors/Fonts/blob/1ce1aff64448bbd15c7418a07d3e645f31683fc0/src/SixLabors.Fonts/TextLayout.cs#L40

I don't see that `TrimEnd` call present in 1.0.0 or further, so perhaps that's been resolved? However, I'm not sure we can update the package. I attempted to update it locally, and now no text draws - perhaps because the API differs greatly and ImageSharp is no longer compatible?

I'm marking this as ready for review without those changes, so it won't look totally accurate when there's a space as in the first image.

Tarcontar (approved):

PR #809, opened 2024-01-26
Records for data types in FileFormats.W3d

charliefoxtwo:
I was looking at enabling nullable reference types in OpenSage.FileFormats.W3d, and found a lot of these immutable data types. I wanted to propose moving these over to [records](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record). Records are purpose-built immutable reference types with a clear contract and I think this would be a good use case for them. 

I've only migrated a few types (including the core `W3dChunk`) in this draft as a sample, as there are a lot of types to migrate so I don't want to go all-in without first making sure it's a direction we want to go in. At a glance it definitely looks possible. As only a few types have been migrated and one of them is a base type for many things, this will not compile as-is. 

One downside I could see is that large migrations like this can muddy up the commit history. Also the phrase "if it ain't broke, don't fix it" comes to mind. This is really just _one way_ of getting to nullable reference types in this project. The other might look something like this (note the `null!` assignment):

```cs
public class MyClass
{
    public Foo Prop1 { get; private set; } = null!;
    
    internal static MyClass Parse(...)
    {
        var myClass = new MyClass();
        var foo = Foo.Parse(...);
        myClass.Prop1 = foo;
        return myClass;
    }
}
```

What I don't like about this method is that it relies on the developer to verify `Prop1` is actually set, which to me defeats the purpose.

Another alternative:

```cs
public class MyClass
{
    public Foo Prop1 { get; }
    
    public MyClass(Foo prop1)
    {
        Prop1 = prop1;
    }
    
    internal static MyClass Parse(...)
    {
        var foo = Foo.Parse(...);
        return new MyClass(foo);
    }
}
```

But now we're just getting closer to record types and at that point, I figure we just lean into them?


charliefoxtwo:
Discussion on discord also floated the idea of using record structs, which should be almost identical in code (one extra word) but may offer performance improvements.

charliefoxtwo:
Upon further investigation, using record structs likely will not be an option as structs are not inheritable in C#

tgjones (approved):
After thinking about this for a few days, I've decided I like it. I like how compact it makes the code. 

I don't know if records are suitable for everywhere in OpenSAGE - e.g. for ModuleData classes, I think I want the properties to be mutable, so they can be updated at runtime via our developer UI. 

But for W3D where we never expect to mutate data, records make sense. Thanks for doing this!

PR #821, opened 2024-01-29
Draw hero reticle on radar

charliefoxtwo:
Not sure if this is a thing in bfme? But if not, it should just be a no-op.

image


charliefoxtwo:
I didn't see HeroReticle anywhere else in the Generals inis so it may be a default parameter or hardcoded there.

Tarcontar (approved):

PR #823, opened 2024-01-30
Fix GLATunnelNetwork GLAStingerSite SpawnBehavior

charliefoxtwo:
Previously, units were spawning at map 0,0 (in the case of the tunnel network) or somewhere along the map edge (in the case of the stinger site).

It appears the issue with the tunnel network was that the productionexitbehavior was specifically commented out in inis with a comment that it falls back to opencontain, so I've implemented support for that here (including following the exit bones and rally points).

For the stinger site, it seems like `_productionExit.GetUnitCreatePoint()` was already in world space, and we transformed it _again_ resulting in some weird coordinates.

@Tarcontar looks like you wrote a lot of this code initially (if a while ago) so I wouldn't mind your eyes on it.

### DIFF
--- a/src/OpenSage.Game/Logic/Object/Behaviors/SpawnBehavior.cs
+++ b/src/OpenSage.Game/Logic/Object/Behaviors/SpawnBehavior.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using OpenSage.Content;
 using OpenSage.Data.Ini;
 
@@ -11,7 +12,8 @@ namespace OpenSage.Logic.Object
 
         private List<GameObject> _spawnedUnits;
         private bool _initial;
-        private IProductionExit _productionExit;
+        private IProductionExit? _productionExit;
+        private OpenContainModule? _openContain;
 
         private bool _unknownBool1;
         private string _templateName;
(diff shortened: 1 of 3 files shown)


src/OpenSage.Game/Logic/Object/Behaviors/SpawnBehavior.cs
```diff
@@ -44,9 +44,25 @@ private void SpawnUnit()
...
+            _productionExit ??= _gameObject.FindBehavior<IProductionExit>();
+
+            if (_productionExit == null)
+            {
+                return false;
+            }
+
             if (_productionExit != null)
             {
-                spawnedObject.SetTranslation(_gameObject.ToWorldspace(_productionExit.GetUnitCreatePoint()));
```
Tarcontar, line 49: This was most likely done for Bfme reasons. Could be the case that we need a dedicated implementation depending on the game

Tarcontar (approved):

src/OpenSage.Game/Logic/Object/Behaviors/SpawnBehavior.cs
charliefoxtwo: Would you be able to test and see if this pr would break bfme? My suspicion is that the GetUnitCreatePoint method changed at some point to return worldspace automatically since that's probably how it was being used everywhere, but I haven't checked.

Tarcontar: @charliefoxtwo sadly BFME is quite broken right now, so there would be a lot of fixes required in order to test things again. In my opinion we should focus on Generals right now, because testing many games at the same time is quite time consuming

Tarcontar (approved):

PR #837, opened 2024-02-04
Implement Special Power timers

charliefoxtwo:
Gets special power countdown timers working and showing in the generals control bar.

### what's working
- powers disabled when countdown not complete
- items which are paused (capture and radar van scan) will not resume countdown until `UnpauseSpecialPowerUpgrade` unpauses them
- using a special power resets the countdown
- shared sync timers are tracked at the player level rather than the object level (think A-10 strike, artillery barrage, rebel ambush - things where having more command centers doesn't give you more abilities)
- `SpecialPowerCreate` resets the timer on creation
- shared sync timers are persisted as in actual generals saves (depends on correct `SpecialPowerType` ordinal values to work properly, see #835)

### what's not yet working
- the special power hotbar does not display the countdown timers currently. It wouldn't be too difficult for the shared items, but I'm not sure how non-shared items should be handled other than simply iterating over every object which feels _very_ inefficient


Tarcontar (approved):

Tarcontar:
@charliefoxtwo please fix the conflicts, otherwise it looks good

PR #852, opened 2024-02-07
Changed the BFME mods and splitted them

Ravo92:
Also i added a Variable from gamedata.ini
Added some Apt-Methods


charliefoxtwo:
I'm not super familiar with BFME in general, but what's the logic for creating a separate project for ROTWK? My understanding of that is that it's an expansion, similar to Generals vs Zero Hour - and at least for those two games, I don't believe we've had a reason to create a new project for Zero Hour thus far. Some files like `AptControlBarSource.cs` look effectively identical to me?

I guess a general explanation of the reasoning here would be nice, especially for someone like me who doesn't typically work with the BFME side of things :)

Ravo92:
I just separated the game into their own project for the sake of keeping it on its own, even when there is some duplicate code.
There are some functions which are only in rotwk and not in bfme 2 itself, but yea, you can kinda throw the merge away and keep only the variables and functions for the APT i added. I just separated it, so i can better organise it.

Tarcontar:
Hi @Ravo92 thank you for your contribution!
I can only speak for myself, but i think we should keep it as unified as possible, since this is already a large project and this would split up things without a real need to.

Ravo92:
> Hi @Ravo92 thank you for your contribution! I can only speak for myself, but i think we should keep it as unified as possible, since this is already a large project and this would split up things without a real need to.

Hi, is it possible for you to throw the project split away and just keep the APT Methods and the variables I added, or do i need to make a new pull request?

charliefoxtwo:
You'll have to edit this PR, or close it and create a new one

Ravo92:
@Tarcontar It was my first time editing a Merge Request, i hope i did alright!

Michaelschnabel-DM (approved):

Michaelschnabel-DM:
@Ravo92 all fine now, thank you! :)

Tarcontar (approved):

PR #860, opened 2024-02-12
Localize build progress and control bar cash

charliefoxtwo:
This uses and formats the proper localized strings for the build progress on screen and in the control bar, as well as the control bar cash display.

image

To do this, https://github.com/adamhewitt627/sprintf.NET is used and imported as a new dependency. It's a rather small and stable library. `LocalizedString` was updated to support formatting as a result.

I'm unsure what font or details should be used for the "Building" message that is displayed visually over a structure being built. I took a guess at some magic numbers that seem to look alright, but if anybody is aware of where these parameters might defined (if they are at all) that would be nice.


Tarcontar (approved):

charliefoxtwo:
Science menu localization+formatting added 
304276474-fd0a5143-1504-4029-ac06-b56ca1e807bd

Tarcontar (approved):

PR #870, opened 2024-02-20
Implement GLA hole behavior

charliefoxtwo:
### `RebuildHoleBehavior` renamed and now extends `UpdateModule`
I know this is called a behavior in the module name, but the data from the sav file seems to parse more properly with UpdateModule as the base, which makes sense as this needs to run every frame to track when the worker should start rebuilding and what to do if the worker dies

### Dead `GameObject`s now only process `SlowDeathBehavior`s
Without this, I noticed that when a worker was killed, as long as they were undergoing their slow death animation they would continue to build the structure. As there's nothing else to do with any of these modules _to my knowledge_ once the unit is dead (and any death-specific behavior should be run in the module's `OnDie` override) I believe this is a safe change.

### Dead buildings with `RebuildHoleExposeDie` spawn a hole on death
This is pretty straightforward, the only catch is that the hole is assigned a max health based on this moduledata, rather than the hole object itself (as all the holes are just "reskins", and the base template has a **lot** of health).

### `RebuildHoleUpdate` implemented
- worker is not selectable
- when the worker dies, the timer to spawn a new one is reset
- if the scaffold dies, the worker is despawned
- if scaffold construction completes, the worker and the hole are both despawned
- TODO: we'll need to determine a way to hide the cancel construction button for a scaffold of this type. These scaffolds have an ObjectStatus of 20 in the sav files, but I'm not sure what that should map to so I don't know if that's relevant here yet.

Obligatory low-quality gif that I feel like has become a trademark of my PRs at this point (at 5x speed)
ezgif-6-d20eb78a52


Tarcontar:
@charliefoxtwo I really appreciate those gifs, keep them coming! :D

Tarcontar (approved):

PR #880, opened 2024-03-04
Parse modules used in particle cannon

charliefoxtwo:
I need to get a real hobby.


charliefoxtwo:
Thinking about this more, I bet all those empty bytes and the identity matrices are because the particleuplinkcannonupdate module supports up to 16 flares and the arrays are all a fixed size. I'll update it accordingly.

Tarcontar (approved):

PR #905, opened 2024-05-28
Fix W3dModelDraw not disposing its particle systems correctly

### LINKED ISSUE #686: Generals: When placing a scaffold and then right-clicking to cancel, particle effects remain
The lower fire is from a canceled barracks, and the upper smoke stack is from a canceled power plant.

Note that construction did not begin for the object so this isn't a die call.

image

In Generals/ZH, the particle effects don't even appear until the object is fully constructed, so we probably just shouldn't be showing these in the ghost or before construction is complete

c001ac1d:
Particle systems now no longer remain after the DrawModule (or Drawable) gets disposed

closes #686

### DIFF
--- a/src/OpenSage.Game/Client/Drawable.cs
+++ b/src/OpenSage.Game/Client/Drawable.cs
@@ -20,6 +20,8 @@ namespace OpenSage.Client
             module.Tag = tag;
             _drawModules.Add(module);
             _tagToModuleLookup.Add(tag, module);
+
+            AddDisposable(module);
         }
 
         private void AddClientUpdateModule(string tag, ClientUpdateModule module)
@@ -367,11 +369,6 @@ namespace OpenSage.Client
                 _shownSubObjects[subObject] = true;
             }
             _hiddenSubObjects.Remove(subObject);
-        }
-
-        public new void Dispose()
-        {
-            Destroy();
         }
 
         internal void Destroy()
--- a/src/OpenSage.Game/Client/Drawable.cs
+++ b/src/OpenSage.Game/Client/Drawable.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
@@ -367,6 +367,11 @@ namespace OpenSage.Client
                 _shownSubObjects[subObject] = true;
             }
             _hiddenSubObjects.Remove(subObject);
+        }
+
+        public new void Dispose()
+        {
+            Destroy();
         }
 
         internal void Destroy()
--- a/src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
+++ b/src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
@@ -366,6 +366,11 @@ namespace OpenSage.Logic.Object
             _activeModelDrawConditionState?.Model?.DrawInspector();
         }
 
+        public override void Dispose()
+        {
+            _activeModelDrawConditionState?.Dispose();
+        }
+
         internal override void Load(StatePersister reader)
         {
             reader.PersistVersion(2);
@@ -428,6 +433,11 @@ namespace OpenSage.Logic.Object
         }
 
         public bool StillActive() => Model.AnimationInstances.Any(x => x.IsPlaying);
+
+        public new void Dispose()
+        {
+            Deactivate();
+        }
 
         public void Deactivate()
         {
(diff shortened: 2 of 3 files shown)


src/OpenSage.Game/Client/Drawable.cs
```diff
@@ -369,6 +369,11 @@ public void ShowSubObjectPermanently(string subObject)
             _hiddenSubObjects.Remove(subObject);
         }
 
+        public new void Dispose()
+        {
+            Destroy();
+        }
+
```
charliefoxtwo, line 376: Instead of doing this, should we maybe add all drawables to the `DisposableBase` via `AddDisposable`, so we can leverage what we're already inheriting?

src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
```diff
@@ -429,6 +434,11 @@ internal sealed class W3dModelDrawConditionState : DisposableBase
 
         public bool StillActive() => Model.AnimationInstances.Any(x => x.IsPlaying);
 
+        public new void Dispose()
+        {
+            Deactivate();
+        }
```
charliefoxtwo, line 440: Perhaps instead of overwriting the `Dispose` method, we could use `DisposableBase.AddDisposeAction`?

Better yet, `AddDisposeAction` could happen within the `ParticleSystem` itself, and then the particle systems here could just be added via `DisposableBase.AddDisposable`?

src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
```diff
@@ -366,6 +366,11 @@ internal override void DrawInspector()
             _activeModelDrawConditionState?.Model?.DrawInspector();
         }
 
+        public override void Dispose()
+        {
+            _activeModelDrawConditionState?.Dispose();
```
charliefoxtwo, line 371: I think this can be added to `DisposableBase.AddDisposable` as well

charliefoxtwo (commented):
Thanks for digging into this!

I think we can leverage `DisposableBase.AddDisposable` (and perhaps `DisposableBase.AddAction` where needed) since all these classes seem to extend `DisposableBase` already. I'm not super familiar with this area of the code though so happy to be proven wrong.

c001ac1d:
@charliefoxtwo i've modified my solution to be a much cleaner one thanks to your review.
> Better yet, AddDisposeAction could happen within the ParticleSystem itself, and then the particle systems here could just be added via DisposableBase.AddDisposable?

i've tested this, unfortunately the game crashes, and i think it's because we dispose the ParticleSystem when the W3dModelDraw gets disposed, however ParticleSystemManager is still trying to render this disposed particle system.
we can just set the particle system's state to Finished, and the manager will automatically take care of freeing the resources afterwards:
https://github.com/OpenSAGE/OpenSAGE/blob/f3f8364885eaccd1c89c3d5f675a0a8b6f4861cb/src/OpenSage.Game/Graphics/ParticleSystems/ParticleSystem.cs#L306-L309

https://github.com/OpenSAGE/OpenSAGE/blob/f3f8364885eaccd1c89c3d5f675a0a8b6f4861cb/src/OpenSage.Game/Graphics/ParticleSystems/ParticleSystemManager.cs#L94-L101

src/OpenSage.Game/Client/Drawable.cs
```diff
@@ -20,6 +20,8 @@ private void AddDrawModule(string tag, DrawModule module)
             module.Tag = tag;
             _drawModules.Add(module);
             _tagToModuleLookup.Add(tag, module);
+
+            AddDisposable(module);
```
charliefoxtwo, line 24: The only concern I have is this is pretty core to like... The whole game. Maybe we've been missing this behavior the whole time though?

For my own sanity I'd like to do some cursory playtests to make sure we're not breaking something weird here (or maybe this will solve a bunch of problems instead, who knows!)

charliefoxtwo (approved):
Just want to test this locally before merging, but thanks again for doing this!

src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
```diff
@@ -425,6 +425,8 @@ internal sealed class W3dModelDrawConditionState : DisposableBase
 
             _particleSystems = particleSystems;
             _context = context;
+
+            AddDisposeAction(() => Deactivate());
```
charliefoxtwo, line 429: Super nitty nitpick that doesn't matter but I'll point out anyway ;)
```suggestion
            AddDisposeAction(Deactivate);
```

charliefoxtwo:
Gave this a test and it looks good - and by proxy also fixes a separate issue where building smoke effects are way overpronounced - probably because we were drawing two at once. Sweet!

**Before**
image
**After**
image

src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
c001ac1d: yeah, looks a bit easier to read

PR #909, opened 2024-05-30
Fixed MappedImageUtility::CreateTexture to handle block compression properly

### LINKED ISSUE #907: Zero Hour: Experience Window background not rendering
### Actual
The zero hour experience window just isn't drawing at all for some reason. It looks like Zero Hour uses images for the background here, and for the life of me I can't figure out why the image isn't appearing. Everything seems to be correct, the image is present in the window and the call is being made to draw it, but nothing seems to appear.
image

c001ac1d:
closes #907 
image


charliefoxtwo:
Nice work!

It does look like the menu is _slightly_ to the right of center (at least relative to the text), but still, awesome!

PR #911, opened 2024-05-30
Fix subobjects not hiding their bones

### LINKED ISSUE #844: Generals: subobjects are not hidden from drawables
image
From left to right:
- marauder tank has 3 multiple barrels, which it only gets with scrap upgrades
- quad cannon also has scrap-upgraded weapon
- radar van has radar mast on the back, which is granted with the Radar Van Scan science researched at the Black Market

It's also quite visible on the shellmap, where humvees have the tow missile upgrade component shown
image

Note in the ini for the humvee that `TurretUp01` is hidden by default, and only shown when the player has the requisite upgrade.
```ini
ConditionState = NONE
  Model = AVHUMMER
  Turret = Turret
  ShowSubObject = Turret
  HideSubObject = TurretUp01 Housecolor03 MuzzleFXUP01
  WeaponFireFXBone = PRIMARY Muzzle
  WeaponMuzzleFlash = PRIMARY MuzzleFX
End

; ...

ConditionState = WEAPONSET_PLAYER_UPGRADE
  Model = AVHUMMER
  Turret = TurretUp01
  HideSubObject = Turret MuzzleFX01
  ShowSubObject = TurretUp01 Housecolor03
  WeaponFireFXBone = PRIMARY MuzzleUp
  WeaponMuzzleFlash = PRIMARY MuzzleFXUp
  WeaponFireFXBone = SECONDARY WeaponB
  WeaponLaunchBone = SECONDARY WeaponB
  WeaponFireFXBone = TERTIARY WeaponB
  WeaponLaunchBone = TERTIARY WeaponB
End
```

c001ac1d:
closes #844

## Before
(taken from #844)
image
## After
image

### DIFF
--- a/src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
+++ b/src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
@@ -272,8 +272,8 @@ namespace OpenSage.Logic.Object
             {
                 var name = subObject.Name;
 
-                if ((subObject.RenderObject.Hidden && !(Drawable.ShownSubObjects?.ContainsKey(name) ?? false))
-                    || (Drawable.HiddenSubObjects?.ContainsKey(name) ?? false))
+                if ((subObject.RenderObject.Hidden && !Drawable.ShownSubObjects.ContainsKey(name))
+                    || Drawable.HiddenSubObjects.ContainsKey(name))
                 {
                     drawState.Model.BoneVisibilities[subObject.Bone.Index] = false;
                     continue;
--- a/src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
+++ b/src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
@@ -85,6 +85,8 @@ namespace OpenSage.Logic.Object
         {
             if (_activeConditionState == conditionState || ShouldWaitForRunningAnimationsToFinish())
             {
+                UpdateBoneVisibilities(_activeConditionState, _activeModelDrawConditionState);
+
                 return;
             }
 
@@ -261,6 +263,26 @@ namespace OpenSage.Logic.Object
             SetActiveAnimationState(bestAnimationState, random);
         }
 
+        private void UpdateBoneVisibilities(ModelConditionState conditionState, W3dModelDrawConditionState drawState)
+        {
+            var model = conditionState.Model?.Value;
+            if (model == null) return;
+
+            foreach (var subObject in model.SubObjects)
+            {
+                var name = subObject.Name;
+
+                if ((subObject.RenderObject.Hidden && !(Drawable.ShownSubObjects?.ContainsKey(name) ?? false))
+                    || (Drawable.HiddenSubObjects?.ContainsKey(name) ?? false))
+                {
+                    drawState.Model.BoneVisibilities[subObject.Bone.Index] = false;
+                    continue;
+                }
+
+                drawState.Model.BoneVisibilities[subObject.Bone.Index] = true;
+            }
+        }
+
         private W3dModelDrawConditionState CreateModelDrawConditionStateInstance(ModelConditionState conditionState, Random random)
         {
             // Load model, fallback to default model.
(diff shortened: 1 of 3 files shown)


c001ac1d:
bonus: we're not checking for hidden subobjects every frame now. the checks happen only when the condition state changes

charliefoxtwo:
Out of curiosity, did you happen to verify that when you purchase an upgrade, the submodule appears? Humvee tow should be the easiest to test this with.

c001ac1d:
> did you happen to verify that when you purchase an upgrade, the submodule appears? Humvee tow should be the easiest to test this with.

left humvee has the state, right one doesn't
image

>I don't recall if we properly update the condition state on an upgrade.

unfortunately we don't, i think we should file an issue if there isn't one already

charliefoxtwo:
Neat!

I actually suspect it will work for like radar vans though.

I seem to recall trying to implement this and running into an issue just because of how the collection of game objects was stored, so I wasn't able to iterate over the collection at the point in time I needed to (since this is a player upgrade which must then be applied to all existing objects of the player). We should file an issue for this if I haven't already.

If you build a new humvee, _hopefully_ it has the upgrade? If not, this is probably also worth filing an issue for.

src/OpenSage.Game/Logic/Object/Draw/W3dModelDraw.cs
```diff
@@ -261,6 +263,26 @@ public override void UpdateConditionState(BitArray<ModelConditionFlag> flags, Ra
...
+        {
+            var model = conditionState.Model?.Value;
+            if (model == null) return;
+
+            foreach (var subObject in model.SubObjects)
+            {
+                var name = subObject.Name;
+
+                if ((subObject.RenderObject.Hidden && !(Drawable.ShownSubObjects?.ContainsKey(name) ?? false))
+                    || (Drawable.HiddenSubObjects?.ContainsKey(name) ?? false))
```
charliefoxtwo, line 276: These should never be null as you've made them autoproperties for the readonly properties beneath them, and those are assigned a new dictionary on object construction
```suggestion
                if ((subObject.RenderObject.Hidden && !Drawable.ShownSubObjects.ContainsKey(name))
                    || Drawable.HiddenSubObjects?.ContainsKey(name))
```

charliefoxtwo: Looking at this a little closer - does this need some extra parentheses? && takes conditional precedence https://stackoverflow.com/questions/1196703/c-sharp-conditional-and-or-precedence

c001ac1d: i may be missing something but doesn't it already have the extra parentheses?
```csharp
(subObject.RenderObject.Hidden && !Drawable.ShownSubObjects.ContainsKey(name))
                    || Drawable.HiddenSubObjects?.ContainsKey(name)
```
boils down to
```
(true && false) || true
```
or if you meant that it DOESN'T need them, well, i think it's just more obvious what's happening when you're reading that with parentheses

c001ac1d: removed the null checks

charliefoxtwo: > i may be missing something but doesn't it already have the extra parentheses?
> 
> ```cs
> (subObject.RenderObject.Hidden && !Drawable.ShownSubObjects.ContainsKey(name))
>                     || Drawable.HiddenSubObjects?.ContainsKey(name)
> ```
> 
> boils down to
> 
> ```
> (true && false) || true
> ```
> 
> or if you meant that it DOESN'T need them, well, i think it's just more obvious what's happening when you're reading that with parentheses

Honestly I have no idea where they _should_ be. I was assuming `a && (b || c)` was desired, but I don't know 🤷

c001ac1d: well I think the desired logic is this:
if the object is marked as hidden in the model, and is not flagged as forcefully shown by the Drawable, then hide it. also if the object is forcefully hidden by the Drawable, then hide it too

c001ac1d: adding to that, we shouldn't have cases where a subobject is both marked hidden and shown in the drawable

charliefoxtwo: Never say never 😆

charliefoxtwo:
We should test those other upgrade cases and file issues as necessary

PR #923, opened 2024-07-28
Implement animation duration scaling for hacker unpacking / packing

tgjones:
#875 implemented `HackInternetAIUpdate`, but it didn't implement scaling the unpacking / packing animations based on the randomized duration. This PR implements that.

### DIFF
--- a/src/OpenSage.Game/Client/Drawable.cs
+++ b/src/OpenSage.Game/Client/Drawable.cs
@@ -385,6 +385,14 @@ namespace OpenSage.Client
             }
         }
 
+        public void SetSupplyBoxesRemaining(float boxPercentage)
+        {
+            foreach (var drawModule in DrawModules)
+            {
+                drawModule.SetSupplyBoxesRemaining(boxPercentage);
+            }
+        }
+
         internal void Destroy()
         {
             foreach (var drawModule in DrawModules)
--- a/src/OpenSage.Game/Client/Drawable.cs
+++ b/src/OpenSage.Game/Client/Drawable.cs
@@ -15,22 +15,6 @@ namespace OpenSage.Client
     {
         private readonly Dictionary<string, ModuleBase> _tagToModuleLookup = new();
 
-        private void AddDrawModule(string tag, DrawModule module)
-        {
-            module.Tag = tag;
-            _drawModules.Add(module);
-            _tagToModuleLookup.Add(tag, module);
-
-            AddDisposable(module);
-        }
-
-        private void AddClientUpdateModule(string tag, ClientUpdateModule module)
-        {
-            module.Tag = tag;
-            _clientUpdateModules.Add(module);
-            _tagToModuleLookup.Add(tag, module);
-        }
-
         private ModuleBase GetModuleByTag(string tag)
         {
             return _tagToModuleLookup[tag];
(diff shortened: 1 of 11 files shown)


src/OpenSage.Game/Client/Drawable.cs
```diff
@@ -52,12 +36,25 @@ private ModuleBase GetModuleByTag(string tag)
...
 
-        // Doing this with a field and a property instead of an auto-property allows us to have a read-only public interface,
-        // while simultaneously supporting fast (non-allocating) iteration when accessing the list within the class.
-        public IReadOnlyList<DrawModule> DrawModules => _drawModules;
-        private readonly List<DrawModule> _drawModules;
+        public ReadOnlySpan<DrawModule> DrawModules
+        {
+            get
+            {
+                if (ModelConditionFlags.BitsChanged)
```
tgjones, line 43: Previously, we only recomputed model condition states when rendering... which made it hard to set e.g. properties of the animation we've just set. So I've changed it to recompute on-demand, and everything that needs the draw modules must go through this property.

src/OpenSage.Game/Logic/Object/Update/AIUpdate/HackInternetAIUpdate.cs
```diff
@@ -97,6 +91,12 @@ public void StartHackingInternet()
                 throw new InvalidStateException();
             }
 
+            GameObject.ModelConditionFlags.Set(ModelConditionFlag.Unpacking, true);
+            GameObject.ModelConditionFlags.Set(ModelConditionFlag.FiringA, false);
+            GameObject.ModelConditionFlags.Set(ModelConditionFlag.Packing, false);
```
tgjones, line 96: Since `Update` isn't called until later, I had to move this to here, so we can then update the animation duration below.

src/OpenSage.Game/Logic/Player.cs
```diff
@@ -883,6 +883,11 @@ public GameObject FindClosestSupplyWarehouse(GameObject supplyGatherer)
             {
                 var supplyWarehouse = _game.GameLogic.GetObjectById(supplyWarehouseId);
 
+                if (supplyWarehouse == null)
```
tgjones, line 886: Unrelated to animation scaling - I just hit a crash here when loading a save file that had supply warehouses in it.

src/OpenSage.Game/Scene3D.cs
```diff
@@ -423,7 +423,7 @@ internal void BuildRenderList(RenderList renderList, Camera camera, in TimeInter
         // This is for drawing 2D elements which depend on the Scene3D, e.g tooltips and health bars.
         internal void Render(DrawingContext2D drawingContext)
         {
-            _scene25D.Draw(drawingContext);
+            _scene25D?.Draw(drawingContext);
```
tgjones, line 426: Unrelated to animation scaling - I hit a crash when opening the developer mode Preview window.

tgjones:
I don't think this is 100% right yet, because I still see it jump to the hacking state before the unpacking animation has completed.

charliefoxtwo (approved):
I tested and it worked great for me, nice!

src/OpenSage.Game/Logic/Object/Update/SupplyWarehouseDockUpdate.cs
```diff
@@ -15,7 +15,7 @@ internal SupplyWarehouseDockUpdate(GameObject gameObject, SupplyWarehouseDockUpd
             _gameObject = gameObject;
             _moduleData = moduleData;
             _currentBoxes = _moduleData.StartingBoxes;
-            _supplyDraw = _gameObject.Drawable.DrawModules.OfType<W3dSupplyDraw>().Single(); // unsure if there is ever more or less than one
+            _supplyDraw = _gameObject.Drawable.DrawModules.ToArray().OfType<W3dSupplyDraw>().Single(); // unsure if there is ever more or less than one
```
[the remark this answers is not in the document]
tgjones: Good point - https://github.com/OpenSAGE/OpenSAGE/pull/923/commits/87e6f1d7e2c02a554ee533f3916947e1ca9e27cf

PR #925, opened 2024-07-30
Fix quadtree returning all items in leaf when no searcher is present

charliefoxtwo:
Added plenty of tests around the quadtree as well to ensure it's behaving properly. Previously if no searcher was present, the entire leaf would be returned, which is not correct as the leaf is not guaranteed to be entirely contained by the querying collider.


tgjones (approved):
Looks great!

charliefoxtwo:
#926 solves this problem better

PR #926, opened 2024-08-01
Replace Quadtree with more accurate implementation

charliefoxtwo:
Supersedes #925 (ported over tests)

## Implementation

After benchmark testing with [Orthogonal Arrays](https://en.wikipedia.org/wiki/Orthogonal_array), the minimum node geometric size and max items per node parameters were tweaked for better performance.

Additionally, an overlap parameter was added. This scales the subgrids by the provided factor, such that each subgrid overlaps its neighbors. This helps prevent hotspots on cell borders from piling up in the parent nodes.

This new implementation persists the existing behavior that two entities whose borders perfectly touch are **not** considered colliding. If that is not the desired behavior we may wish to address that separately.

This also switches to using an interface in the method signatures instead of the class name itself, partially because that made it easier for me to flip back and forth between testing and partially so that if someone else ever wants to burn this quadtree with fire and replace it with their implementation that's even better, it will be easier for them as well :)

### Insert
I would have expected the performance here to be similar, so it's interesting the new quadtree is so much faster.
| Method         | MaxItemSize | InsertedItems | Mean       | Error     | StdDev    | Ratio |
|--------------- |------------ |-------------- |-----------:|----------:|----------:|------:|
| InsertQuadtree | 15          | 5000          | 839.493 us | 8.9584 us | 8.3797 us | 1.000 |
| InsertNewTree   | 15          | 5000          |   3.346 us | 0.0241 us | 0.0226 us | 0.004 |

### Update
I'm not quite sure why the old quadtree updates were so slow. This benchmark also isn't actually recording the performance of a single update, but rather multiple.
| Method         | TotalItems | MovingItems | Mean         | Error      | StdDev     | Ratio |
|--------------- |----------- |------------ |-------------:|-----------:|-----------:|------:|
| UpdateQuadtree | 15000      | 10000       | 3,205.934 ms | 11.2854 ms | 10.0043 ms | 1.000 |
| UpdateNewTree   | 15000      | 10000       |     4.007 ms |  0.0544 ms |  0.0455 ms | 0.001 |

### Query
Here it's no surprise that the new quadtree is slower - many items were missed in the previous implementation, so it's not really an apples-to-apples comparison.
| Method            | ItemCount | ItemSize | AreaSideLength | Mean       | Error    | StdDev  | Ratio | RatioSD |
|------------------ |---------- |--------- |--------------- |-----------:|---------:|--------:|------:|--------:|
| QueryAreaQuadtree | 3000      | 10       | 100            |   406.1 ns |  2.62 ns | 2.05 ns |  1.00 |    0.00 |
| QueryAreaNewTree   | 3000      | 10       | 100            | 1,634.2 ns | 10.37 ns | 9.19 ns |  4.02 |    0.04 |

### DIFF
--- a/src/OpenSage.Benchmarks.DataStructures/BenchQuadtreeItem.cs
+++ b/src/OpenSage.Benchmarks.DataStructures/BenchQuadtreeItem.cs
@@ -27,11 +27,10 @@ namespace OpenSage.Benchmarks.DataStructures
 
         public static BenchQuadtreeItem Generate(int id, RectangleF bounds, SizeF maxSize, Random random)
         {
-            var x = (float) random.NextDouble() * bounds.Width - bounds.Left;
-            var y = (float) random.NextDouble() * bounds.Height - bounds.Top;
-
-            var width = MathF.Min(bounds.Right - x, (float) (random.NextDouble() * maxSize.Width));
-            var height = MathF.Min(bounds.Bottom - y, (float) (random.NextDouble() * maxSize.Height));
+            var x = (float)(bounds.X + random.NextDouble() * bounds.Width);
+            var y = (float)(bounds.Y + random.NextDouble() * bounds.Height);
+            var width = (float)(random.NextDouble() * Math.Min(bounds.Width - x, maxSize.Width));
+            var height = (float)(random.NextDouble() * Math.Min(bounds.Height - y, maxSize.Height));
 
             return new BenchQuadtreeItem(id, new RectangleF(x, y, width, height));
         }
--- a/src/OpenSage.Game.Tests/DataStructures/QuadtreeTests.cs
+++ b/src/OpenSage.Game.Tests/DataStructures/QuadtreeTests.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
 using OpenSage.DataStructures;
@@ -29,9 +30,30 @@ namespace OpenSage.Tests.DataStructures
                 Colliders = new List<Collider> { collider };
             }
 
+            // implementation copied from GameObject
             public bool CollidesWith(ICollidable other, bool twoDimensional)
             {
-                throw new System.NotImplementedException();
+                if (RoughCollider == null || other.RoughCollider == null)
+                {
+                    return false;
+                }
+
+                if (!RoughCollider.Intersects(other.RoughCollider, twoDimensional))
+                {
+                    return false;
+                }
+
+                foreach (var collider in Colliders)
+                {
+                    foreach (var otherCollider in other.Colliders)
+                    {
+                        if (collider.Intersects(otherCollider, twoDimensional))
+                        {
+                            return true;
+                        }
+                    }
+                }
+                return false;
             }
         }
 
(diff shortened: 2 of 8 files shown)


src/OpenSage.Benchmarks.DataStructures/BenchQuadtreeItem.cs
```diff
@@ -27,11 +27,10 @@ public BenchQuadtreeItem(int id, RectangleF bounds)
...
         {
-            var x = (float) random.NextDouble() * bounds.Width - bounds.Left;
-            var y = (float) random.NextDouble() * bounds.Height - bounds.Top;
-
-            var width = MathF.Min(bounds.Right - x, (float) (random.NextDouble() * maxSize.Width));
-            var height = MathF.Min(bounds.Bottom - y, (float) (random.NextDouble() * maxSize.Height));
+            var x = (float)(bounds.X + random.NextDouble() * bounds.Width);
+            var y = (float)(bounds.Y + random.NextDouble() * bounds.Height);
+            var width = (float)(random.NextDouble() * Math.Min(bounds.Width - x, maxSize.Width));
+            var height = (float)(random.NextDouble() * Math.Min(bounds.Height - y, maxSize.Height));
```
charliefoxtwo, line 33: I replaced this (specifically width/height) to attempt to better distribute the sizes of the items generated. Previously this would generate any size and then cap it to max size, which means that a majority of items would be max size.

src/OpenSage.Game.Tests/DataStructures/QuadtreeTests.cs
```diff
@@ -135,5 +157,293 @@ public void InsertAndRemoveNonExistant()
...
+                        continue;
+                    }
+
+                    if (!collisions.ContainsKey(item2.Id))
+                    {
+                        collisions[item2.Id] = [];
+                    }
+
+                    collisions[item1.Id].Add(item2.Id);
+                    collisions[item2.Id].Add(item1.Id);
```
charliefoxtwo, line 438: If nothing else, we should make sure this is correct. As long as this is correct, then if the fuzzing passes we can feel confident that our quadtree implementation is correct.

src/OpenSage.Game/DataStructures/ICollidable.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+using System.Numerics;
+using OpenSage.Logic.Object;
+
+namespace OpenSage.DataStructures;
+
+public interface ICollidable
```
charliefoxtwo, line 7: This is just split into a separate file (previously was nested in Quadtree)

src/OpenSage.Game/DataStructures/QuadtreeNode.cs
```diff
@@ -0,0 +1,196 @@
...
+            }
+
+            if (CollidesForReal(data, collider, searcher, twoDimensional, hasToCollideWithSearcher))
+            {
+                yield return data;
+            }
+        }
+    }
+
+    private static bool CollidesForReal(T item, Collider collider, T? searcher, bool twoDimensional, bool hasToCollideWithSearcher)
```
charliefoxtwo, line 68: I just realized I never went back and renamed this - open to suggestions for a better method name here 😆

src/OpenSage.Game/DataStructures/Quadtree.cs
```diff
@@ -1,266 +1,60 @@
...
-        List<Collider> Colliders { get; }
-        Vector3 Translation { get; }
+        return _rootNode.Remove(item);
+    }
 
-        bool CollidesWith(ICollidable other, bool twoDimensional);
+    public bool Update(in T item)
+    {
+        Remove(item);
+        return Insert(item);
```
charliefoxtwo, line 58: This was just copied over from the existing implementation, but there's probably room for optimization here. If we can determine that the node the item is in still fully captures the item, then we don't actually need to do anything here.

src/OpenSage.Game/DataStructures/QuadtreeNode.cs
tgjones: Can you describe in one sentence what the method does? I've read the code but I'm not 100% sure - is it a precise collision test?

tgjones (approved):

tgjones:
This is awesome!

Often when the benchmark results are wildly different from expectations, it's worth double checking that the benchmarks actually measure what you think they do 😄 But assuming they do, then this looks like a very nice speedup on insert and update, with an expected slowdown for query since it's returning more data.

src/OpenSage.Game/DataStructures/QuadtreeNode.cs
charliefoxtwo: Ah, that's it - `CollidesPrecise` should do the trick.

Yeah, this was ported over from the original implementation, but basically the quadtree does most of its checks on the GameObject's `RoughCollider` (so, an AABB), and then if that matches then this precise version is called which checks all the _actual_ collidables against each other.

charliefoxtwo:
More benchmarks here couldn't hurt (remove isn't benchmarked at all currently), that's for sure. Since the new quadtree has an identical public signature to the old one, they used the exact same code in the benchmark methods, so I'm pretty comfortable saying they _should have_ been testing the same exact same thing.

I bet this could be further optimized for even better performance as well - but I'm not super experienced in that area.

Issue #879, opened 2024-03-01
HackInternetAIUpdate does not transition to proper exit states


charliefoxtwo:
As there's still a lot to do for state machines, #875 didn't handle many AI state machine transitions when giving an order to a hacker that is hacking. However, as these sorts of orders probably aren't given to hackers that often, it may be easy for this missing behavior to go unnoticed for quite some time, so I'm filing this issue so we remember.

Behaviors that are covered:
- Stop

Behaviors that are sort of covered:
- Move (doesn't use the proper move state machine, but that's not implemented yet)
- Garrison (since garrison doesn't use a state machine currently)

Behaviors that are not covered:
- Attack Move (accessible when selected with another non-hacker unit)
- Defend
- ???

PR #927, opened 2024-08-01
Move hacking behavior to state classes

tgjones:
As a prerequisite to #879, I wanted to make `State` subclasses contain not just data, but also behavior. 

We've done this already with our [WeaponStateMachine](https://github.com/OpenSAGE/OpenSAGE/blob/master/src/OpenSage.Game/Logic/Object/Weapon/WeaponStates/WeaponStateMachine.cs), and it's pretty standard in any case. See [Super Simple State Machine](https://keithmaggio.wordpress.com/code/supersimplestatemachine/) for another example. 

Having `State` subclasses contain behavior and handle transitions to the next state should provide a nice framework for implementing more of these.

### DIFF
--- a/src/OpenSage.Game/Logic/AI/AIStates/StopHackingInternetState.cs
+++ b/src/OpenSage.Game/Logic/AI/AIStates/StopHackingInternetState.cs
@@ -4,13 +4,54 @@ namespace OpenSage.Logic.AI.AIStates
 {
     internal sealed class StopHackingInternetState : State
     {
-        public LogicFrameSpan FramesUntilFinishedPacking;
+        public const uint Id = 1002;
+
+        private readonly GameObject _gameObject;
+
+        private LogicFrameSpan _framesUntilFinishedPacking;
+
+        public StopHackingInternetState(GameObject gameObject)
+        {
+            _gameObject = gameObject;
+        }
+
+        public override void OnEnter()
+        {
+            _gameObject.ModelConditionFlags.Set(ModelConditionFlag.Unpacking, false);
+            _gameObject.ModelConditionFlags.Set(ModelConditionFlag.FiringA, false);
+            _gameObject.ModelConditionFlags.Set(ModelConditionFlag.Packing, true);
+
+            _gameObject.GameContext.AudioSystem.PlayAudioEvent(_gameObject, _gameObject.Definition.UnitSpecificSounds.UnitPack?.Value);
+
+            var aiUpdate = (HackInternetAIUpdate)_gameObject.AIUpdate;
+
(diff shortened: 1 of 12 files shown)


src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -1476,6 +1476,8 @@ public void Persist(StatePersister reader)
             reader.PersistUInt32(ref teamId);
             Team = GameContext.Game.TeamFactory.FindTeamById(teamId);
 
+            Owner = Team.Template.Owner;
```
tgjones, line 1479: This is unrelated to the PR. I just found that game objects are assigned to the wrong owner when loading a `.sav` file.

src/OpenSage.Game/Logic/AI/AIStates/StopHackingInternetState.cs
```diff
@@ -4,13 +4,54 @@ namespace OpenSage.Logic.AI.AIStates
...
+        {
+            _gameObject.ModelConditionFlags.Set(ModelConditionFlag.Unpacking, false);
+            _gameObject.ModelConditionFlags.Set(ModelConditionFlag.FiringA, false);
+            _gameObject.ModelConditionFlags.Set(ModelConditionFlag.Packing, true);
+
+            _gameObject.GameContext.AudioSystem.PlayAudioEvent(_gameObject, _gameObject.Definition.UnitSpecificSounds.UnitPack?.Value);
+
+            var aiUpdate = (HackInternetAIUpdate)_gameObject.AIUpdate;
+
+            var frames = aiUpdate.GetVariableFrames(aiUpdate.ModuleData.PackTime);
```
charliefoxtwo, line 28: It would be nice if we could pass this in (or perhaps just expose a Set method or similar? really it would be nice if we could _construct_ this state with this parameter) so we don't have to access the AIUpdate object from here, and then `ModuleData` and `GetVariableFrames` could remain private. However, given the way we currently set states, that could be a more complex task.

The same goes for StartHackingInternetState as well. Perhaps some sort of `transitionTimer` or similar could be added and utilized for other modules as well?

charliefoxtwo (approved):
Nice! I like that we can keep the State data within their respective classes now.

src/OpenSage.Game/Logic/AI/AIStates/StopHackingInternetState.cs
tgjones: > Perhaps some sort of transitionTimer or similar could be added and utilized for other modules as well?

That could be nice, but I think I'd prefer to implement a few more modules and then see what common behaviour we want to extract into helper classes. I just don't know yet what we'll need.

PR #931, opened 2024-08-02
Update slaved module capabilities

charliefoxtwo:
UpgradeDie is now properly implemented (though removing applied upgrades isn't yet supported - we'll still have to figure out how we want to tackle that one), and additional parts of SlavedUpdate are now persisted (though the module still needs to be fixed up to be fully compatible with generals saves).

### child death
The upgrade is no longer considered purchased (though is still disabled as it's considered triggered), and the spy drone upgrade is now purchasable.
OpenSage Launcher_VkZiLPBkmJ

### parent death
When the parent dies, the child dies too.
OpenSage Launcher_ttsWQR89jF

### DIFF
--- a/src/OpenSage.Game/Logic/Object/Behaviors/SpawnBehavior.cs
+++ b/src/OpenSage.Game/Logic/Object/Behaviors/SpawnBehavior.cs
@@ -38,11 +38,9 @@ namespace OpenSage.Logic.Object
             var spawnedObject = _gameObject.GameContext.GameLogic.CreateObject(_moduleData.SpawnTemplate.Value, _gameObject.Owner);
             _spawnedUnits.Add(spawnedObject);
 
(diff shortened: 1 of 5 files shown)


charliefoxtwo:
@Tarcontar @feliwir it looks like you two did a lot of work on `SlavedUpdate` in the past - would appreciate if you're able to take a look at this and:
1) ensure I didn't break any existing behavior
2) see if you have any ideas how the existing repair logic could fit into the persisted state

Tarcontar:
The changes look good to me, cant tell if anything existing might be broken and though and I am way too out of touch with the project to really be of any help integrating the repair logic.

charliefoxtwo:
> The changes look good to me, cant tell if anything existing might be broken and though and I am way too out of touch with the project to really be of any help integrating the repair logic.

Maybe it's time to get back into it 👀

tgjones (approved):
LGTM 😄

PR #935, opened 2024-08-02
Fix units not sinking/disappearing after death

charliefoxtwo:
## Before
Unit and rubble persist indefinitely.
OpenSage Launcher_lnArPBSgDI

## After
Unit and rubble sink and are deleted according to `LifetimeUpdate` and `SlowDeathBehavior`
OpenSage Launcher_mRhyKxLQjh

### DIFF
--- a/src/OpenSage.Game/Logic/Object/ObjectCreationList.cs
+++ b/src/OpenSage.Game/Logic/Object/ObjectCreationList.cs
@@ -137,6 +137,7 @@ namespace OpenSage.Logic.Object
             debrisObject.LifeTime = context.LogicFrame + lifeTime;
 
             debrisObject.UpdateTransform(context.GameObject.Translation + Offset, context.GameObject.Rotation);
+            debrisObject.Die(DeathType.Normal);
 
             // Model
             var w3dDebrisDraw = (W3dDebrisDraw) debrisObject.Drawable.DrawModules[0];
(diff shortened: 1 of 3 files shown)


src/OpenSage.Game/Logic/Object/ObjectCreationList.cs
```diff
@@ -137,6 +137,7 @@ internal override List<GameObject> Execute(BehaviorUpdateContext context, Vector
             debrisObject.LifeTime = context.LogicFrame + lifeTime;
 
             debrisObject.UpdateTransform(context.GameObject.Translation + Offset, context.GameObject.Rotation);
+            debrisObject.Die(DeathType.Normal);
```
charliefoxtwo, line 140: `GenericDebris` has no `LifetimeUpdate`, so it needs to be told to die right away it seems?

src/OpenSage.Game/Logic/Object/Update/LifetimeUpdate.cs
```diff
@@ -33,6 +33,7 @@ internal override void Update(BehaviorUpdateContext context)
             if (context.LogicFrame >= _frameToDie)
             {
                 _gameObject.Die(_moduleData.DeathType);
+                _frameToDie = LogicFrame.MaxValue;
```
charliefoxtwo, line 36: Without this, `_gameObject.Die` gets called repeatedly.

tgjones (approved):

PR #938, opened 2024-08-05
Implement .sav parsing for Zero Hour

tgjones:
I wanted to be able to open Zero Hour `.sav` files, but I found a few missing bits and pieces. This PR implements enough things to be able to open simple `.sav` files with a small selection of objects.

### DIFF
--- a/src/OpenSage.Game/Client/Drawable.cs
+++ b/src/OpenSage.Game/Client/Drawable.cs
@@ -403,7 +403,7 @@ namespace OpenSage.Client
 
         public void Persist(StatePersister reader)
         {
-            reader.PersistVersion(5);
+            var version = reader.PersistVersion(7);
 
             var id = ID;
             reader.PersistUInt32(ref id);
@@ -552,6 +552,18 @@ namespace OpenSage.Client
             {
                 throw new InvalidStateException();
             }
+
+            if (version >= 7)
+            {
+                var unknownBool4 = true;
+                reader.PersistBoolean(ref unknownBool4);
+                if (!unknownBool4)
+                {
+                    throw new InvalidStateException();
+                }
+
+                reader.SkipUnknownBytes(1);
+            }
         }
 
         private void PersistModules(StatePersister reader)
@@ -623,5 +635,7 @@ namespace OpenSage.Client
         NationalismVehicle = 4, // exhordeb_up.dds
         Crate = 5, // exjunkcrate.dds
         None = 6,
+        Unknown1 = 7,
+        Unknown2 = 8,
     }
 }
--- a/src/OpenSage.Game/Logic/Object/Behaviors/GenerateMinefieldBehavior.cs
+++ b/src/OpenSage.Game/Logic/Object/Behaviors/GenerateMinefieldBehavior.cs
@@ -33,6 +33,11 @@ namespace OpenSage.Logic.Object
             reader.PersistBoolean(ref _unknown);
 
             reader.SkipUnknownBytes(13);
+
+            if (reader.SageGame == SageGame.CncGeneralsZeroHour)
+            {
+                reader.SkipUnknownBytes(2);
+            }
         }
     }
 
--- a/src/OpenSage.Game/Logic/Object/Behaviors/GenerateMinefieldBehavior.cs
+++ b/src/OpenSage.Game/Logic/Object/Behaviors/GenerateMinefieldBehavior.cs
@@ -34,7 +34,7 @@ namespace OpenSage.Logic.Object
 
             reader.SkipUnknownBytes(13);
 
-            if (reader.SageGame == SageGame.CncGeneralsZeroHour)
+            if (reader.SageGame >= SageGame.CncGeneralsZeroHour)
             {
                 reader.SkipUnknownBytes(2);
             }
--- a/src/OpenSage.Game/Logic/Object/Body/ActiveBody.cs
+++ b/src/OpenSage.Game/Logic/Object/Body/ActiveBody.cs
@@ -111,7 +111,7 @@ namespace OpenSage.Logic.Object
             var maxHealth2 = _maxHealth;
             reader.PersistSingle(ref maxHealth2);
 
-            if (reader.SageGame == SageGame.CncGeneralsZeroHour)
+            if (reader.SageGame >= SageGame.CncGeneralsZeroHour)
             {
                 var maxHealth3 = _maxHealth;
                 reader.PersistSingle(ref maxHealth3);
--- a/src/OpenSage.Game/Logic/Object/Body/ActiveBody.cs
+++ b/src/OpenSage.Game/Logic/Object/Body/ActiveBody.cs
@@ -110,9 +110,15 @@ namespace OpenSage.Logic.Object
 
             var maxHealth2 = _maxHealth;
             reader.PersistSingle(ref maxHealth2);
-            if (_maxHealth != maxHealth2)
+
+            if (reader.SageGame == SageGame.CncGeneralsZeroHour)
             {
-                throw new InvalidStateException();
+                var maxHealth3 = _maxHealth;
+                reader.PersistSingle(ref maxHealth3);
+                if (maxHealth3 != maxHealth2)
+                {
+                    throw new InvalidStateException();
+                }
             }
 
             reader.PersistEnum(ref _damageType);
--- a/src/OpenSage.Game/Logic/Object/GameObject.cs
+++ b/src/OpenSage.Game/Logic/Object/GameObject.cs
@@ -504,6 +504,13 @@ namespace OpenSage.Logic.Object
             }
 
             // TODO: This shouldn't be added to all objects. I don't know what the rule is.
+            if (_gameContext.Game.SageGame >= SageGame.CncGeneralsZeroHour)
+            {
+                AddBehavior("ModuleTag_StatusDamageHelper", new StatusDamageHelper());
+                AddBehavior("ModuleTag_SubdualDamageHelper", new SubdualDamageHelper());
+            }
+
+            // TODO: This shouldn't be added to all objects. I don't know what the rule is.
             // Maybe KindOf = CAN_ATTACK ?
             AddBehavior("ModuleTag_DefectionHelper", new ObjectDefectionHelper());
 
@@ -525,8 +532,6 @@ namespace OpenSage.Logic.Object
             // TODO: This shouldn't be added to all objects. I don't know what the rule is.
             if (_gameContext.Game.SageGame >= SageGame.CncGeneralsZeroHour)
             {
-                AddBehavior("ModuleTag_StatusDamageHelper", new StatusDamageHelper());
-                AddBehavior("ModuleTag_SubdualDamageHelper", new SubdualDamageHelper());
                 AddBehavior("ModuleTag_TempWeaponBonusHelper", new TempWeaponBonusHelper());
             }
 
--- a/src/OpenSage.Game/Logic/Object/GameObject.cs
+++ b/src/OpenSage.Game/Logic/Object/GameObject.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Frozen;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -173,6 +173,7 @@ namespace OpenSage.Logic.Object
         private GameObjectUnknownFlags _unknownFlags;
         private readonly ShroudReveal _shroudRevealSomething1 = new();
         private readonly ShroudReveal _shroudRevealSomething2 = new();
+        private readonly ShroudReveal _shroudRevealSomething3 = new();
         private float _visionRange;
         private float _shroudClearingRange;
 
@@ -519,6 +520,14 @@ namespace OpenSage.Logic.Object
             {
                 // this was added in bfme and is not present in generals or zero hour
                 AddBehavior("ModuleTag_ExperienceHelper", new ExperienceUpdate(this));
+            }
+
+            // TODO: This shouldn't be added to all objects. I don't know what the rule is.
+            if (_gameContext.Game.SageGame >= SageGame.CncGeneralsZeroHour)
+            {
+                AddBehavior("ModuleTag_StatusDamageHelper", new StatusDamageHelper());
+                AddBehavior("ModuleTag_SubdualDamageHelper", new SubdualDamageHelper());
+                AddBehavior("ModuleTag_TempWeaponBonusHelper", new TempWeaponBonusHelper());
             }
 
             foreach (var behaviorDataContainer in objectDefinition.Behaviors.Values)
@@ -1455,7 +1464,7 @@ namespace OpenSage.Logic.Object
 
         public void Persist(StatePersister reader)
         {
-            reader.PersistVersion(7);
+            var version = reader.PersistVersion(9);
 
             var id = ID;
             reader.PersistObjectID(ref id, "ObjectId");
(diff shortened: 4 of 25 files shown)


src/OpenSage.Game/Client/Drawable.cs
```diff
@@ -623,5 +635,7 @@ public enum ObjectDecalType
         NationalismVehicle = 4, // exhordeb_up.dds
         Crate = 5, // exjunkcrate.dds
         None = 6,
+        Unknown1 = 7,
+        Unknown2 = 8,
```
charliefoxtwo, line 639: I also came across `Unknown2`, which an `AmericaVehicleDozer` had? I found this interesting since I'm not aware of any visible decal drawn for a USA dozer...

src/OpenSage.Game/Client/Drawable.cs
```diff
@@ -403,7 +403,7 @@ internal void Destroy()
 
         public void Persist(StatePersister reader)
         {
-            reader.PersistVersion(5);
+            var version = reader.PersistVersion(7);
```
charliefoxtwo, line 406: I think we need to persist this don't we? That way when we save, we save the correct version? Or is the plan to save everything with the latest version?

src/OpenSage.Game/Logic/Object/Body/ActiveBody.cs
```diff
@@ -110,9 +110,15 @@ internal override void Load(StatePersister reader)
 
             var maxHealth2 = _maxHealth;
             reader.PersistSingle(ref maxHealth2);
-            if (_maxHealth != maxHealth2)
+
+            if (reader.SageGame == SageGame.CncGeneralsZeroHour)
```
charliefoxtwo, line 114: for behavior like this, when I did it I went with `if (reader.SageGame is not SageGame.CncGenerals)` since I assumed all future games would use this new version - but I also didn't verify.

src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -521,6 +522,14 @@ void AddBehavior(string tag, BehaviorModule behavior)
                 AddBehavior("ModuleTag_ExperienceHelper", new ExperienceUpdate(this));
             }
 
+            // TODO: This shouldn't be added to all objects. I don't know what the rule is.
+            if (_gameContext.Game.SageGame >= SageGame.CncGeneralsZeroHour)
+            {
+                AddBehavior("ModuleTag_StatusDamageHelper", new StatusDamageHelper());
+                AddBehavior("ModuleTag_SubdualDamageHelper", new SubdualDamageHelper());
```
charliefoxtwo, line 529: I don't know why, but I had these two added before `ModuleTag_DefectionHelper`. Maybe that was the order they were in? Or maybe I was drunk when I wrote that code 🤷

src/OpenSage.Game/Logic/Object/Helpers/TempWeaponBonusHelper.cs
```diff
@@ -0,0 +1,22 @@
...
+        private int _unknownInt;
+
+        internal override void Load(StatePersister reader)
+        {
+            reader.PersistVersion(1);
+
+            base.Load(reader);
+
+            reader.PersistInt32(ref _unknownInt);
+            if (_unknownInt != -1)
```
charliefoxtwo, line 14: I had `if (_unknownInt != 0 && _unknownInt != -1)`, so 0 may have occurred for some reason in one of my saves.

src/OpenSage.Game/Logic/Object/Helpers/StatusDamageHelper.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace OpenSage.Logic.Object.Helpers
+{
+    internal sealed class StatusDamageHelper : ObjectHelperModule
+    {
+        internal override void Load(StatePersister reader)
+        {
+            reader.PersistVersion(1);
+
+            base.Load(reader);
```
charliefoxtwo, line 9: The pattern we've followed elsewhere I believe is:
```cs
reader.BeginObject("Base");
base.Load(reader);
reader.EndObject();
```
I think that helps nest the json properly when writing saves.

src/OpenSage.Game/Logic/Object/ObjectStatus.cs
```diff
@@ -57,6 +57,9 @@ public enum ObjectStatus
         [IniEnum("STATUS_RIDER7")]
         StatusRider7,
 
+        [IniEnum("IS_BRAKING")]
+        IsBraking,
```
charliefoxtwo, line 61: I don't think this can go here - at least not without moving some other things.

Somewhere in the gameobject, ObjectStatus is stored as a uint32 flags enum. We currently have 32 items in this enum, so that's a separate issue, but also this goes between `Hijacked = 9` and `Sold = 18` and the gap between those is already filled with other enums (though their values haven't been confirmed).

https://github.com/OpenSAGE/OpenSAGE/blob/e6cdafabe06f2f7dcc2e430da0fc8782560354fa/src/OpenSage.Game/Logic/Object/GameObject.cs#L1493

If this is added here, then `CanAttack` and `Sold` would share the same ordinal. We should probably take a closer look at this at some point. It might be nice to try to create more of these states to validate their values.

src/OpenSage.Game/Logic/Player.cs
```diff
@@ -653,7 +658,12 @@ public void Persist(StatePersister reader)
                     persister.PersistBooleanValue(ref item);
                 });
 
-            reader.SkipUnknownBytes(70);
+            reader.SkipUnknownBytes(4);
+
+            if (reader.SageGame == SageGame.CncGenerals)
+            {
+                reader.SkipUnknownBytes(66);
+            }
```
charliefoxtwo, line 666: I ran into this same thing and it was so confusing to me

src/OpenSage.Game/Network/SkirmishGameSettings.cs
```diff
@@ -118,6 +121,30 @@ public void Persist(StatePersister reader)
             reader.PersistUInt32(ref _mapFileSize);
             reader.PersistUInt32(ref _unknownInt4);
             reader.PersistUInt32(ref _unknownInt5);
+
+            if (version >= 4)
+            {
+                ushort unknown6 = 0;
+                reader.PersistUInt16(ref unknown6);
```
charliefoxtwo, line 128: I believe the first byte here is for the `Limit Superweapons` behavior - I wasn't able to figure out the second byte though.

src/OpenSage.Game/Network/SkirmishGameSettings.cs
```diff
@@ -118,6 +121,30 @@ public void Persist(StatePersister reader)
...
+            else
+            {
+                // TODO: Set default starting cash.
+            }
+        }
+    }
+
+    internal struct Money : IPersistableObject
+    {
+        public uint Amount;
```
charliefoxtwo, line 141: This solves the above todo and is what I did personally (though I didn't have this in a dedicated object, didn't realize it was so good catch), up to you if this is the route you want to go though.
```suggestion
        public uint Amount = 10000u;
```

charliefoxtwo (commented):
The biggest concern I have is with not storing the version locally - that means we'll just always be saving things as the latest version. Just want to make sure that's an intentional design decision.

Other times I think we're checking if SageGame == ZeroHour, but we probably want `is not SageGame.Generals` (or `>= SageGame.ZeroHour` as you've done elsewhere in this PR)

We should also figure out what to do about that ObjectStatus issue at some point, but that's not necessarily a new issue here.

Two other things I came across:

In `PartitionCellManager` I had to add this, so it's interesting that you didn't have to! I wonder what was going on there
```cs
if (_game.SageGame is not SageGame.CncGenerals)
{
    // zero hour saves seem to have one extra cell
    // todo: does this affect bfme as well?
    _numCellsX++;
    _numCellsY++;
}
```

and in `TeamFactory` I had to delete the `if (count != _teamTemplates.Count)` check - they didn't match up either.

Curious why those only affected me, but maybe it was particular to the map I was using.

src/OpenSage.Game/Client/Drawable.cs
tgjones: I was thinking we'd save everything with the latest version, yes. The only reason to save with older versions would be "export" a save file that's compatible with original versions of older games, and I don't think there's really a need for that. In fact I mainly see importing .sav files as a development aid for us, to help us learn about the original games.

I _did_ save the version in the W3D parser because I thought maybe we could use it as the basis of a .w3d converter between SAGE games, but I don't see the same value for save files.

What do you think?

src/OpenSage.Game/Logic/Object/ObjectStatus.cs
tgjones: That's actually why I put it near the end, close to the values that were added in BFME and later games. It's only older save files (specifically Generals) that persist the status as a uint, so the numeric values of values added in later games don't matter. But you're right, I didn't put it quite low enough - I'll move it below `AirborneTarget`.

src/OpenSage.Game/Logic/Object/Body/ActiveBody.cs
tgjones: Good catch - https://github.com/OpenSAGE/OpenSAGE/commit/b54a4fc795ff4f83fb2043871ff4e6747b1a56d2

src/OpenSage.Game/Logic/Object/GameObject.cs
tgjones: https://github.com/OpenSAGE/OpenSAGE/commit/fcca96a81535183ca3a38cce4f67068449c8b89e

src/OpenSage.Game/Logic/Object/Helpers/TempWeaponBonusHelper.cs
tgjones: Interesting. I'll leave it like it is for now but that's good to know. I'll try to upload some test .sav files to give us some regression tests for this code.

src/OpenSage.Game/Logic/Object/Helpers/StatusDamageHelper.cs
tgjones: Thanks, I missed that one - https://github.com/OpenSAGE/OpenSAGE/commit/c0ecc72697f272d7eaadcb356c3053614de592c3

src/OpenSage.Game/Logic/Object/ObjectStatus.cs
tgjones: I've moved the values around - https://github.com/OpenSAGE/OpenSAGE/commit/64c9a55423ae005bf1e5332f832ab06bc5050df9

src/OpenSage.Game/Network/SkirmishGameSettings.cs
tgjones: Nice - https://github.com/OpenSAGE/OpenSAGE/commit/9f689a9b4d4b8e3add2bac8aae0e0668c9734a52

src/OpenSage.Game/Network/SkirmishGameSettings.cs
tgjones: Ah thanks. I've done it a slightly different way, using the value parsed from `GameData.ini` - https://github.com/OpenSAGE/OpenSAGE/commit/4d08c8071f6fa0cf281bda99ffb0a0e7236dd656

src/OpenSage.Game/Client/Drawable.cs
charliefoxtwo: I'm totally fine with saving everything as the latest version - but is that always possible? For example, there are cases where going from Generals to Zero Hour bytes are _removed_, which seems like something that could get us if maybe in Generals certain data was needed for a given behavior but in Zero Hour that behavior was removed from the engine and thus no longer supported.

I agree that we don't need to worry about saving a game in OpenSAGE and then loading it in vanilla Generals - I think the level of engineering effort required for something like that would be astronomical compared to the value gained.

src/OpenSage.Game/Network/SkirmishGameSettings.cs
charliefoxtwo: ooh, this is way better, good call

src/OpenSage.Game/Logic/Object/ObjectStatus.cs
charliefoxtwo: Still conflicts with `InsideGarrison = 21` 😅 

That should be the last item with a known ID

src/OpenSage.Game/Logic/Object/Behaviors/GenerateMinefieldBehavior.cs
```diff
@@ -33,6 +33,11 @@ internal override void Load(StatePersister reader)
             reader.PersistBoolean(ref _unknown);
 
             reader.SkipUnknownBytes(13);
+
+            if (reader.SageGame == SageGame.CncGeneralsZeroHour)
```
charliefoxtwo, line 37:
```suggestion
            if (reader.SageGame >= SageGame.CncGeneralsZeroHour)
```
I guess there's not any way to know for _sure_ if this is correct, but we can assume?

src/OpenSage.Game/Logic/Object/Helpers/SubdualDamageHelper.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace OpenSage.Logic.Object.Helpers
+{
+    internal sealed class SubdualDamageHelper : ObjectHelperModule
+    {
+        internal override void Load(StatePersister reader)
+        {
+            reader.PersistVersion(1);
+
+            base.Load(reader);
```
charliefoxtwo, line 9:
```suggestion
            reader.BeginObject("Base");
            base.Load(reader);
            reader.EndObject();
```

src/OpenSage.Game/Logic/Object/Helpers/TempWeaponBonusHelper.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    internal sealed class TempWeaponBonusHelper : ObjectHelperModule
+    {
+        private int _unknownInt;
+
+        internal override void Load(StatePersister reader)
+        {
+            reader.PersistVersion(1);
+
+            base.Load(reader);
```
charliefoxtwo, line 11:
```suggestion
            reader.BeginObject("Base");
            base.Load(reader);
            reader.EndObject();
```

src/OpenSage.Game/Logic/Object/Update/OCLUpdate.cs
```diff
@@ -3,6 +3,24 @@
...
 {
+    public sealed class OCLUpdate : UpdateModule
+    {
+        private bool _unknownBool;
+
+        internal override void Load(StatePersister reader)
+        {
+            reader.PersistVersion(1);
+
+            base.Load(reader);
```
charliefoxtwo, line 14:
```suggestion
            reader.BeginObject("Base");
            base.Load(reader);
            reader.EndObject();
```

src/OpenSage.Game/Logic/Object/Upgrade/ArmorUpgrade.cs
```diff
@@ -2,6 +2,24 @@
...
+            : base(gameObject, moduleData)
+        {
+            _moduleData = moduleData;
+        }
+
+        internal override void Load(StatePersister reader)
+        {
+            reader.PersistVersion(1);
+
+            base.Load(reader);
```
charliefoxtwo, line 19:
```suggestion
            reader.BeginObject("Base");
            base.Load(reader);
            reader.EndObject();
```

charliefoxtwo (commented):
Added suggestions for more missing begin/end objects around `base.Load(reader);` - I only commented on the one the first time since I didn't want to spam with comments if it was intentional :)

src/OpenSage.Game/Logic/Object/Behaviors/GenerateMinefieldBehavior.cs
tgjones: I can't remember whether any subsequent games used minefields 😄 If not, then I guess it's academic anyway.

charliefoxtwo (approved):

PR #943, opened 2024-08-09
Award cash on building sell

charliefoxtwo:
Previously this assumed that the default was null, but the default is actually 0


tgjones (approved):

tgjones:
Nice!

PR #952, opened 2024-08-11
Persist rally points

charliefoxtwo:
There may be other behaviors which persist rally points, but these are the ones I was able to find. Initially I thought about just having `IProductionExit` implement `IHasRallyPoint`, but `SpawnPointProductionExitUpdate` also implements `IProductionExit` and a cursory investigation doesn't reveal any obvious place that it would store the rally point.


tgjones (approved):
Nice to share this parsing logic 👍

PR #953, opened 2024-08-11
Parse more AI data

tgjones:
I've made a bunch of .sav files with a Chinook doing various actions - repairing, evacuating troops, combat-dropping troops, taking off, landing, exiting map after a reinforcement script action, etc. These .sav files contained a variety of new state data, and the parsing code for those states is in this PR.

The `AIStateMachine._stateSomethingId` field appears to be an "override state", which is a state that temporarily overrides the current state. When the override state has finished, it goes back to the original current state. I found this by telling a Nuke Cannon to fire at an object, and then while it was deployed, I told it to move to a new location. While it was packing up, I made a save, and found that the state machine's state is already set to 1 (`MoveTowardsState`), but the override state is set to 41 (I've called it `DoNothingState`). When packing has finished, the override state is removed and it continues with state `1`.

Separately, I've found that `UnknownStateData` is probably a queued "move to different AI state" command. When a Chinook is being repaired at an airfield, if you tell it to move to a location, then the state will immediately update to the `1001` (takeoff) state, and `ChinookAIUpdate._queuedCommand` will be set.

I don't know why there are two mechanisms for essentially the same thing.


src/OpenSage.Game/Logic/AI/AIStates/ChinookCombatDropState.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            ResetRopes(numRopes);
+        }
+
+        reader.BeginArray("Ropes");
+        foreach (var rope in _ropes)
+        {
+            reader.PersistObject(rope);
+        }
+        reader.EndArray();
```
charliefoxtwo, line 44: test this, but I think it should work
```suggestion
        reader.PersistListWithUInt32Count(_ropes, (StatePersister persister, ref Rope item) =>
        {
            item ??= new Rope();
            persister.PersistObjectValue(item);
        });
```

src/OpenSage.Game/Logic/AI/AIStates/ChinookExitMapState.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace OpenSage.Logic.AI.AIStates;
+
+internal sealed class ChinookExitMapState : State
+{
+    public override void Persist(StatePersister reader) { }
```
charliefoxtwo, line 5: not even a version?

src/OpenSage.Game/Logic/Object/Behaviors/ParkingPlaceBehavior.cs
```diff
@@ -245,13 +248,29 @@ internal override void Load(StatePersister reader)
...
+            reader.PersistListWithByteCount(_healingData, (StatePersister persister, ref HealingData item) =>
             {
-                throw new InvalidStateException();
+                persister.PersistObjectValue(ref item);
+            });
+
+            reader.SkipUnknownBytes(13);
+
+            // This is the default 0x3FFFFFFFu, unless we are healing something,
+            // in which case it's CurrentFrame + a small number of frames.
```
charliefoxtwo, line 261: Looks like the healing interval (based on the moduledata) is per second - any chance this is the "next heal frame" or something?

charliefoxtwo (approved):
Nice!

src/OpenSage.Game/Logic/AI/AIStates/ChinookExitMapState.cs
tgjones: Nope, not even a version. Usually there's a version even when there's no data to persist, but... not always, it seems.

src/OpenSage.Game/Logic/AI/AIStates/ChinookCombatDropState.cs
tgjones: Thanks! https://github.com/OpenSAGE/OpenSAGE/pull/953/commits/daabe241f40c986df4012dabdbc9f0289655aebc

src/OpenSage.Game/Logic/Object/Behaviors/ParkingPlaceBehavior.cs
tgjones: Good idea, that's quite likely - https://github.com/OpenSAGE/OpenSAGE/pull/953/commits/5948e1de5420a022201e253c259ef094a84bf2ba

PR #957, opened 2024-08-12
Implement QueueProductionExitUpdate

### LINKED ISSUE #948: Generals: Building China Red Guard should produce two units
It's not clear to me if these two units are spawned with a slight delay, or spawned at the same time slightly apart from each other

## Vanilla Generals
game dat_P23hTYcDJz

## OpenSAGE
OpenSage Launcher_bcu5cL73I2

charliefoxtwo:
Includes a fairly sizeable refactor of ProductionUpdate and ProductionJob as well, updating the classes so most of the core state information is persisted.

Fixes #948

Some of the BFME stuff has been commented out as it no longer fits the proper implementation. We'll also still need to investigate scaling production times (for example, USA/China low power or GLA with power)

OpenSage Launcher_cziofLQwdF

### DIFF
--- a/src/OpenSage.Game/Logic/Object/BehaviorModule.cs
+++ b/src/OpenSage.Game/Logic/Object/BehaviorModule.cs
@@ -168,6 +168,26 @@ namespace OpenSage.Logic.Object
             return !(left == right);
         }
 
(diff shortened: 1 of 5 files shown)


tgjones (approved):
Great work!

PR #959, opened 2024-08-18
Add Chinook AI .sav files

tgjones:
This PR adds some .sav files with a bunch of different Chinook AI states.

I also looked at some diffs of the .json files produced by `OpenSage.Tools.Sav2Json` and made a few fixes / tweaks based on that.


src/OpenSage.Game/Logic/Object/Contain/OpenContain.cs
```diff
@@ -21,7 +21,7 @@ public abstract class OpenContainModule : UpdateModule, IHasRallyPoint
         private uint _unknownFrame1;
         private uint _unknownFrame2;
         private uint _unknownInt2;
-        protected BitArray<ModelConditionFlag> ModelConditionFlags = new();
+        private BitArray<ModelConditionFlag> _cachedModelConditionFlags = new();
```
charliefoxtwo, line 24: Is this ever being updated now? I see that it was being set in some subclasses and those usages were removed. I don't know how much we _need_ it either or if it should ever differ from the GameObject.

charliefoxtwo (approved):

src/OpenSage.Game/Logic/Object/Contain/OpenContain.cs
tgjones: It doesn’t make sense to have as a duplicate of the object’s model condition flags. My guess is that on every update, it’s used to determine whether the model condition flags have changed. If they have, something is triggered. So I don’t think we should set it from the subclasses. Instead, whenever we figure out what the update and triggering logic needs to be, we can implement that.

PR #960, opened 2024-08-18
Use central package management

tgjones:
[Central package management](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management) is the modern way to manage package versions. Let's use it.


charliefoxtwo (approved):
This is neat, I didn't know about this!

PR #967, opened 2024-08-24
Add debug function to persister to log segment bytes

charliefoxtwo:
I've found myself using this function a _lot_ when testing. Maybe someone can point me to a better way, but my workflow has been:
- create a sav file with the behavior I'm interested in introspecting
- replace the `Load` function of the behavior with `LogToSegmentEnd()`
- run Sav2Json

If this would benefit others, I'd like to commit this that way I don't have to keep stashing and popping it 😆 

Output might look something like this. It's also easy to copy/paste something like this into HexEd.it for further processing.
```
[22:23:22.3900][Trace] Skirmish game status is now Started
[22:23:22.4205][Info] Set active condition state for AmericaVehicleDozer
01 00 00 00 00 00 00 00 00 00 00 
[22:23:22.4424][Info] Weapon DozerMineDisarmingWeapon on game object  transitioning to state Inactive
[22:23:22.4424][Info] Set active condition state for AmericaCommandCenter
```

### DIFF
--- a/src/OpenSage.Game/StatePersister.cs
+++ b/src/OpenSage.Game/StatePersister.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Numerics;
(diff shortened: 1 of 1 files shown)


tgjones (approved):
I like it 👍

PR #971, opened 2024-08-29
Track damage dealer

charliefoxtwo:
Also adds tests for the updated DamageData objects.

Tracking the damage dealer is needed for tracking who kills what for both veterancy and for things like salvage crates (salvage crates are only generated when a unit is killed by an enemy SALVAGER).


src/OpenSage.Game/Logic/Object/Weapon/WeaponTemplate.cs
```diff
@@ -591,13 +591,20 @@ public enum WeaponPrefireType
         PerPosition
     }
 
+    // items with explicit indexes have been verified against saves generated by Generals
     public enum DeathType
     {
         [IniEnum("NORMAL")]
-        Normal,
+        Normal = 0,
```
tgjones, line 598: The XSDs for C&C 3 have this for `DeathType`:

``` xml
<xs:simpleType name="DeathType">
	<xs:restriction base="xs:string">
		<xs:enumeration value="NONE" />
		<xs:enumeration value="NORMAL" />
		<xs:enumeration value="CRUSHED" />
		<xs:enumeration value="BURNED" />
		<xs:enumeration value="EXPLODED" />
		<xs:enumeration value="POISONED" />
		<xs:enumeration value="TOPPLED" />
		<xs:enumeration value="FLOODED" />
		<xs:enumeration value="SUICIDED" />
		<xs:enumeration value="LASERED" />
		<xs:enumeration value="DETONATED" />

		<xs:enumeration value="SPLATTED" />
		<xs:enumeration value="POISONED_BETA" />

		<xs:enumeration value="EXTRA_2" />
		<xs:enumeration value="EXTRA_3" />
		<xs:enumeration value="EXTRA_4" />
		<xs:enumeration value="EXTRA_5" />
		<xs:enumeration value="EXTRA_6" />
		<xs:enumeration value="EXTRA_7" />
		<xs:enumeration value="EXTRA_8" />

		<xs:enumeration value="KNOCKBACK" />
		<xs:enumeration value="SUPERNATURAL" />
		<xs:enumeration value="FADED" />
		<xs:enumeration value="SLAUGHTERED" />
		<xs:enumeration value="CATALYST" />

		<!-- special case for all on -->
		<xs:enumeration value="ALL" />
	</xs:restriction>
</xs:simpleType>
```

I'm assuming the enum indices match the internal enum definition.

I wonder if `NONE` and `NORMAL` were switched around between Generals and C&C3, so what you've called `UnknownHealing` might be `NONE`.

`EXPLODED = 4` agrees with the XSD enum index.

Looking at the value at index 3, it's `BURNED`, so perhaps that's what `UnknownContainsDeath` is?

tgjones (approved):
Awesome!

src/OpenSage.Game/Logic/Object/Weapon/WeaponTemplate.cs
charliefoxtwo: hm - I can test burned with a dragon tank 🤔

charliefoxtwo: It is in fact burned. I'm very intrigued as to why objects in a container had that death type applied.

charliefoxtwo:
@tgjones I did some more testing, and you were right - the order of that tib wars xsd enum seems to be correct, with the exception (as you noted) that "Normal" and "None" are reversed. I tested None by setting a rifleman's weapon's death type to none and then having it attack something. It did in fact record the death type as 1, so that seems to be correct. Interesting that it was reversed in later games - not sure how we'd resolve that if we were to add those games to OpenSAGE as initially planned, but at this rate that's probably not something we'll have to worry about until we're retired lol

src/OpenSage.Game/Logic/Object/Weapon/WeaponTemplate.cs
tgjones: I think that will be controlled by this:

https://github.com/OpenSAGE/OpenSAGE/blob/cddd87650428f653a7a6911c893bf25f6d6ed843/src/OpenSage.Game/Logic/Object/Contain/TransportContain.cs#L199

charliefoxtwo: That seems to only be explicitly set to false for gla bikes in zero hour, so I guess it's otherwise true by default

tgjones (approved):

tgjones:
> not sure how we'd resolve that if we were to add those games to OpenSAGE as initially planned, but at this rate that's probably not something we'll have to worry about until we're retired lol

Yeah, I think we can cross that bridge if we ever come to it 😄

PR #972, opened 2024-08-30
Improve .sav file parsing

tgjones:
This PR has a few tweaks for parsing `.sav` files, including giving names to a few more fields related to pathing and physics. I used a test map with a single Humvee, and a script that moved that Humvee only in the X direction, to work this out.


charliefoxtwo (approved):
With some of this PhysicsBehavior decoding, I wonder if this might be a decent opportunity to start adding persister tests - that way as we start to investigate more of the fields, we have a sort of recording of previously-seen values.

tgjones:
Yes, I was thinking that too - although I was thinking more along the lines of general `PhysicsBehavior` tests, e.g. add a force and check that the field values match what's expected, based on previously-seen .sav values.

PR #973, opened 2024-08-31
Add PhysicsBehavior tests

tgjones:


src/OpenSage.Game/StatePersister.cs
```diff
@@ -370,8 +370,9 @@ public override void LogToSegmentEnd()
             Console.WriteLine(segment.Name);
             for (var i = _binaryReader.BaseStream.Position; i < segment.End; i++)
             {
+                Console.Write("0x");
                 Console.Write(_binaryReader.ReadByte().ToString("x2"));
-                Console.Write(" ");
+                Console.Write(", ");
```
tgjones, line 375: I made these changes so I could copy and paste the output directly into a state persister test.

charliefoxtwo: I initially didn't have this stuff so I could copy it into HexEd.it, but then I found out you can copy with the 0x prefix and the comma separators, so I'm totally fine with this

charliefoxtwo (approved):

PR #976, opened 2024-08-31
Implement CreateCrateDie

### LINKED ISSUE #975: CreateCrateDie

charliefoxtwo:
Fixes #975

Creating a crate would normally be disabled with a null killer, but I've disabled that check for these tests. Additionally, the crate normally wouldn't be visible for a non-GLA player, but that hasn't been implemented yet.

Changes also needed to be made to run DeletionUpdate for "dead" objects, so that crates are deleted per their behavior definition.

OpenSage Launcher_Aw06lqjCL6


tgjones (approved):
Looks good 👍

PR #977, opened 2024-08-31
Add W3dScienceModelDraw

charliefoxtwo:
Not sure if this is the right way to do this, but it does work.

This is only used by Salvage Crates in Generals, so that only GLA players see them. As implemented salvage crates are still generated as expected in CreateCrateDie (see #976) but are not displayed for non-GLA players.

It's also unclear how this works for observers, since an observer (who doesn't have their view set to a non-GLA player) should be able to see them too, but observers are probably a deep in the future problem.

### DIFF
--- a/src/OpenSage.Game/Logic/Object/Draw/W3dScienceModelDraw.cs
+++ b/src/OpenSage.Game/Logic/Object/Draw/W3dScienceModelDraw.cs
@@ -1,20 +1,40 @@
-﻿using OpenSage.Client;
+﻿#nullable enable
+
+using OpenSage.Client;
+using OpenSage.Content;
 using OpenSage.Data.Ini;
 
 namespace OpenSage.Logic.Object
 {
     public sealed class W3dScienceModelDraw : W3dModelDraw
     {
+        private readonly W3dScienceModelDrawModuleData _data;
+        private readonly Drawable _drawable;
+        private readonly GameContext _context;
+
         internal W3dScienceModelDraw(W3dScienceModelDrawModuleData data, Drawable drawable, GameContext context)
             : base(data, drawable, context)
         {
+            _data = data;
+            _drawable = drawable;
+            _context = context;
         }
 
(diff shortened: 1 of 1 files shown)


src/OpenSage.Game/Logic/Object/Draw/W3dScienceModelDraw.cs
```diff
@@ -1,20 +1,40 @@
...
         internal override void Load(StatePersister reader)
         {
             reader.PersistVersion(1);
 
+            reader.BeginObject("Base");
             base.Load(reader);
+            reader.EndObject();
+        }
+
+        internal override void Update(in TimeInterval gameTime)
```
tgjones, line 32: What about overriding `BuildRenderList` and not calling the base method instead? I don't feel strongly about it though.

tgjones (approved):
I don't know the best way to let replay observers see the crate, but yes I guess we can punt on that for now 😄

src/OpenSage.Game/Logic/Object/Draw/W3dScienceModelDraw.cs
charliefoxtwo: I don't have a lot of knowledge in this area, so I'm very open to suggestions. If you think that would be better, I'm all for it.

I do wonder if whatever solution we go with should make sure objects become unhidden when the player acquires the requisite science. I don't remember how SCIENCE_GLA is granted - is that from a cc? Or do you have to have the army at start? I'm wondering if you capture a GLA cc, does scrap become visible?

I could also see someone editing the inis to make something like a scrap collection science, but that's a bit down the line for us to worry about I think.

charliefoxtwo: Just checked, and it looks like `SCIENCE_GLA` is an intrinsic science granted as part of a player template, which indicates to me that if you capture a cc you don't get to start collecting scrap.

PR #978, opened 2024-08-31
Extract GameObjectCollection to interface, push down to GameLogic

charliefoxtwo:
## What does this even do?
`GameObjectCollection` was a convenience class which just called through to `GameLogic` and allowed us to expose effectively a subset of `GameLogic`'s methods. This just adds those methods as an interface on `GameLogic`, and converts all calls to calling `GameLogic` directly. Some methods were renamed as the naming differed slightly between `GameLogic` and `GameObjectCollection`. Ideally, functionally nothing changes and everything continues to run as before.

## Ok so if it does nothing why the hell would you do this?
As an interface, it's much easier to construct objects for tests without needing a full `GameContext` or `GameLogic` (as the previous class did). It's also nice to reduce the number of ways to get to things so that the intended access path is cleaner. Many objects take in a full `GameContext` or `GameLogic` when they really only need a `GameObjectCollection`. This PR doesn't aim to clean all _those_ up as well, however - that can be done on a case-by-case basis as we start improving test coverage and decomposing objects into pieces that are easier to construct.


tgjones (approved):
Very nice cleanup!

PR #979, opened 2024-09-01
Add PhysicsBehavior.Update test

tgjones:
I don't love the things I had to do add this test 😄 But the test itself is _roughly_ the kind of thing I'd like to write.

### DIFF
--- a/src/OpenSage.Game/GameContext.cs
+++ b/src/OpenSage.Game/GameContext.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using OpenSage.Audio;
 using OpenSage.Client;
+using OpenSage.Content;
 using OpenSage.Content.Loaders;
 using OpenSage.DataStructures;
 using OpenSage.Graphics.ParticleSystems;
@@ -19,11 +20,13 @@ namespace OpenSage
         public readonly Navigation.Navigation Navigation;
         public readonly Radar Radar;
 
-        public Game Game => Scene3D.Game;
+        public readonly IGame Game;
 
-        internal GameLogic GameLogic => Scene3D.Game.GameLogic;
+        internal GameLogic GameLogic => Game.GameLogic;
 
-        internal GameClient GameClient => Scene3D.Game.GameClient;
+        internal GameClient GameClient => Game.GameClient;
+
+        public AssetStore AssetStore => AssetLoadContext.AssetStore;
 
         public readonly Random Random = new Random();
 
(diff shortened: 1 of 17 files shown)


src/OpenSage.Game/Terrain/TerrainLogic.cs
```diff
@@ -1,7 +1,16 @@
-﻿namespace OpenSage.Terrain
+﻿using OpenSage.Data.Map;
+
+namespace OpenSage.Terrain
 {
     public sealed class TerrainLogic : IPersistableObject
     {
+        public HeightMap HeightMap { get; private set; }
```
charliefoxtwo, line 7: It feels weird to me that both TerrainLogic and Terrain are now holding separate references to a HeightMap - is this intended?

src/OpenSage.Game/GameContext.cs
```diff
@@ -51,7 +54,8 @@ internal GameContext(
             Navigation.Navigation navigation,
             Radar radar,
             IQuadtree<GameObject> quadtree,
-            Scene3D scene)
+            Scene3D scene,
```
charliefoxtwo, line 57: Van we remove Scene3D, or are there too many external usages still?

charliefoxtwo (approved):
Not following what the TerrainLogic changes are for, but the rest of this looks good!

src/OpenSage.Game/Terrain/TerrainLogic.cs
tgjones: I think it’s the same object instance shared by both? I needed access to it from PhysicsBehavior, but I can’t create a Terrain instance since that has a load of graphics stuff. And I also didn’t want to remove Terrain.HeightMap because it had a lot of references. At some point it should be refactored though.

charliefoxtwo: Sorry by separate references I meant separate references to the same object - just feels odd Terrain can't call into TerrainLogic or vv

charliefoxtwo: It's not a big deal though - I think the testability here is the big win!

src/OpenSage.Game/GameContext.cs
tgjones: Still too many usages. It's something I want to refactor though.

src/OpenSage.Game/Terrain/TerrainLogic.cs
tgjones: I guess I could pass `TerrainLogic` into `Terrain` - but ideally we should move things in `Terrain` into either `TerrainLogic` or `TerrainVisual`.

PR #980, opened 2024-09-01
Separate Radar and Radar drawing responsibilities

charliefoxtwo:
Includes unit tests for Radar with most of the fields identified.


src/OpenSage.Game/StatePersister.cs
```diff
@@ -367,15 +369,22 @@ public override void SkipUnknownBytes(int numBytes)
...
-            for (var i = _binaryReader.BaseStream.Position; i < segment.End; i++)
+            LogBytes(segment.End - _binaryReader.BaseStream.Position, segment.Name);
+            Segments.Push(segment);
+        }
+
+
+        /// <summary>
+        /// Only to be used for local debugging and should not be present in committed code.
+        /// </summary>
+        public override void LogBytes(long numBytes, string source = "")
```
charliefoxtwo, line 380: `LogToSegmentEnd()` only works when we have a segment - which we don't always. It works great for Behaviors, but not for nested objects. This is a more versatile logging method which allows for dumping data of specific objects.

src/OpenSage.Game/StatePersister.cs
```diff
@@ -830,6 +839,33 @@ public static void PersistColorRgba(this StatePersister persister, ref ColorRgba
             }
         }
 
+        /// <summary>
+        /// This is not a typo - in some places color is stored as bgr, not rgb
+        /// </summary>
+        public static void PersistColorBgra(this StatePersister persister, ref ColorRgba value, [CallerArgumentExpression("value")] string name = "")
```
charliefoxtwo, line 845: Yes, really

src/OpenSage.Game/RadarDrawUtil.cs
charliefoxtwo, line 1: This file is effectively just the draw functionality extracted from Radar. Nothing new here has been added (well, except for a single null check).

charliefoxtwo (commented):
I left RadarDrawUtil colocated with Radar for now. I could see a situation in the future where that drawing responsibility is perhaps moved out of OpenSAGE.Game altogether, but I didn't see a reason to solve that problem immediately.

src/OpenSage.Game/Radar.cs
```diff
@@ -1,94 +1,40 @@
...
     {
-        private readonly Scene3D _scene;
-        private readonly Texture _miniMapTexture;
+        public RadarItemCollection VisibleItems => _visibleItems;
+        public RadarItemCollection HiddenItems => _hiddenItems;
 
-        private readonly RadarItemCollection _visibleItems;
-        private readonly RadarItemCollection _hiddenItems;
+        // fixed-length array
+        public IReadOnlyList<RadarEvent> RadarEvents => _radarEvents;
```
tgjones, line 18: I think ReadOnlySpan would be more optimal here, as long as the call sites can work with that. ReadOnlySpan allows for iteration without memory allocation.

tgjones (approved):
Lovely!

src/OpenSage.Game/Radar.cs
charliefoxtwo: Great catch!

charliefoxtwo: Updated, let me know if that's what you had in mind. From what I can tell it has to be new'd up since it can't be stored as a field unless it's a ref struct (I mean you don't _have_ to explicitly new it up, but if you don't it will via an implicit cast so I figured best to make it explicit)

tgjones: That’s what I meant, nice. I’m fine with the implicit cast though, I think it’s a common-enough pattern for arrays / ReadOnlySpan that it doesn’t need to be explicitly called out.

charliefoxtwo: Updated to an implicit cast. I'll trust your judgement of what's more common in this area :)

PR #981, opened 2024-09-02
Remove instances of calling through GameObject to access GameContext

charliefoxtwo:
### If this isn't the direction we want to go in, that's fine - we can close this

I've noticed a lot of instances of calling through GameObject to get to GameContext when there's already a GameContext nearby. This removes the public exposure of GameContext from GameObject to encourage the use of GameContext from wherever it already is. Most of these usages are in the Modules, where module creation already has a GameContext fed in. There are a couple instances outside of the modules though, and I've refactored those to either take a GameContext in the method/constructor or just the dependency of GameContext it actually needs.

If we want to continue accessing GameContext via GameObject though, I won't be offended :)


tgjones (approved):
I don't feel too strongly about this either way, but I do like consistency, so thanks for doing this.

PR #985, opened 2024-09-03
Implement FireWeaponWhenDeadBehavior

### LINKED ISSUE #984: FireWeaponWhenDeadBehavior

tgjones:
Fixes #984.

### DIFF
--- a/src/OpenSage.Game/Logic/Object/Behaviors/FireWeaponWhenDeadBehavior.cs
+++ b/src/OpenSage.Game/Logic/Object/Behaviors/FireWeaponWhenDeadBehavior.cs
@@ -10,13 +10,15 @@ namespace OpenSage.Logic.Object;
 public sealed class FireWeaponWhenDeadBehavior : BehaviorModule, IUpgradeableModule
 {
     private readonly GameObject _gameObject;
+    private readonly GameContext _context;
     private readonly FireWeaponWhenDeadBehaviorModuleData _moduleData;
 
     internal UpgradeLogic UpgradeLogic { get; }
 
-    internal FireWeaponWhenDeadBehavior(GameObject gameObject, FireWeaponWhenDeadBehaviorModuleData moduleData)
+    internal FireWeaponWhenDeadBehavior(GameObject gameObject, GameContext context, FireWeaponWhenDeadBehaviorModuleData moduleData)
     {
         _gameObject = gameObject;
+        _context = context;
         _moduleData = moduleData;
         UpgradeLogic = new UpgradeLogic(moduleData.UpgradeData, OnUpgrade);
     }
@@ -49,7 +51,7 @@ public sealed class FireWeaponWhenDeadBehavior : BehaviorModule, IUpgradeableMod
                 _gameObject,
                 deathWeaponTemplate,
                 WeaponSlot.Primary,
-                _gameObject.GameContext);
+                _context);
 
             deathWeapon.Fire();
         }
@@ -92,6 +94,6 @@ public sealed class FireWeaponWhenDeadBehaviorModuleData : UpgradeModuleData
 
     internal override BehaviorModule CreateModule(GameObject gameObject, GameContext context)
     {
-        return new FireWeaponWhenDeadBehavior(gameObject, this);
+        return new FireWeaponWhenDeadBehavior(gameObject, context, this);
     }
 }
--- a/src/OpenSage.Game/Logic/Object/Behaviors/FireWeaponWhenDeadBehavior.cs
+++ b/src/OpenSage.Game/Logic/Object/Behaviors/FireWeaponWhenDeadBehavior.cs
@@ -1,68 +1,97 @@
-﻿using System.Numerics;
+﻿#nullable enable
+
+using System.Numerics;
+using OpenSage.Content;
 using OpenSage.Data.Ini;
 using OpenSage.Mathematics;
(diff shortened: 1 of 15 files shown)


src/OpenSage.Game/Logic/Object/ObjectStatus.cs
```diff
@@ -66,6 +66,15 @@ public enum ObjectStatus
...
 
+        [IniEnum("IS_ATTACKING")]
+        IsAttacking,
+
+        [IniEnum("IS_FIRING_WEAPON")]
+        IsFiringWeapon,
+
+        [IniEnum("IS_AIMING_WEAPON")]
+        IsAimingWeapon,
+
```
tgjones, line 77: Unrelated to this PR, I just happened to notice these while loading ZH .sav files.

src/OpenSage.Game/Logic/Object/Update/AIUpdate/MissileAIUpdate.cs
```diff
@@ -90,7 +90,7 @@ internal override void Update(BehaviorUpdateContext context)
 
         internal override void Load(StatePersister reader)
         {
-            reader.PersistVersion(4);
+            var version = reader.PersistVersion(6);
```
tgjones, line 93: Unrelated to this PR.

src/OpenSage.Game/Logic/Object/Behaviors/FireWeaponWhenDeadBehavior.cs
```diff
@@ -1,68 +1,97 @@
...
+        .Concat(new IniParseTable<FireWeaponWhenDeadBehaviorModuleData>
         {
-            return new FireWeaponWhenDeadBehavior(this);
-        }
+            { "DeathWeapon", (parser, x) => x.DeathWeapon = parser.ParseWeaponTemplateReference() },
+            { "DelayTime", (parser, x) => x.DelayTime = parser.ParseInteger() },
+            { "WeaponOffset", (parser, x) => x.WeaponOffset = parser.ParseVector3() },
+        });
+
+    public DieLogicData DieData { get; } = new();
```
charliefoxtwo, line 83: Is it possible to just push this all the way down to the base class? An easy way to test would be to edit the inis and apply one of the properties to some other DieModule type which doesn't have them documented and see if they work. I noticed with the CrateBehavior stuff I was working on that a lot of the data actually worked in a base class and so the properties and logic could be pushed down.

I guess even if the original game doesn't accommodate them, there's no reason we couldn't. Might clean up the code too!

charliefoxtwo (approved):
Some questions and suggestions I'll leave to your discretion

src/OpenSage.Game/Logic/Object/Behaviors/FireWeaponWhenDeadBehavior.cs
```diff
@@ -1,68 +1,97 @@
...
+
+public sealed class FireWeaponWhenDeadBehavior : BehaviorModule, IUpgradeableModule
 {
-    public sealed class FireWeaponWhenDeadBehavior : BehaviorModule, IUpgradeableModule
+    private readonly GameObject _gameObject;
+    private readonly FireWeaponWhenDeadBehaviorModuleData _moduleData;
+
+    internal UpgradeLogic UpgradeLogic { get; }
+
+    internal FireWeaponWhenDeadBehavior(GameObject gameObject, FireWeaponWhenDeadBehaviorModuleData moduleData)
```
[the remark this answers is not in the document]
tgjones: https://github.com/OpenSAGE/OpenSAGE/pull/985/commits/468e021aaacff452151eddf2d281655ea0398220

src/OpenSage.Game/Logic/Object/Behaviors/FireWeaponWhenDeadBehavior.cs
```diff
@@ -1,68 +1,97 @@
...
-        public bool CanUpgrade(UpgradeSet existingUpgrades) => _upgradeLogic.CanUpgrade(existingUpgrades);
+    public void TryUpgrade(UpgradeSet completedUpgrades) => UpgradeLogic.TryUpgrade(completedUpgrades);
 
-        public void TryUpgrade(UpgradeSet completedUpgrades) => _upgradeLogic.TryUpgrade(completedUpgrades);
+    private void OnUpgrade() { }
 
-        private void OnUpgrade()
+    internal override void OnDie(BehaviorUpdateContext context, DeathType deathType, BitArray<ObjectStatus> status)
+    {
+        if (!_moduleData.DieData.IsApplicable(deathType, status))
```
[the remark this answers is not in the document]
tgjones: It is already in `DieModule`, so modules that inherit from that don't need to call it - they just override the `Die` method. 

But `FireWeaponWhenDeadBehavior` doesn't inherit from `DieModule`. It has both upgrade and die functionality. I've adopted the naming convention that when a module like this one ends in `Behavior`, I inherit from neither `DieModule` nor `UpgradeModule`. Instead I inherit from `BehaviorModule` and include upgrade and die functionality via reusable helper classes, `UpgradeLogic` and `DieData`.

Before this PR, multiple modules (e.g. those inheriting from `DieModule`, and those that don't but need die functionality) duplicated the code for determining whether a death is applicable. I've cleaned that up in this PR to share that logic.

PR #987, opened 2024-09-05
Update ParkingPlaceBehavior and add tests

charliefoxtwo:
#983

All bytes of ParkingPlaceBehavior in Generals are now fully parsed. Zero Hour hasn't been tested yet.

Aircraft still have a bit to go before we get them properly working - namely JetAIUpdate (#986) needs some progress. I've parsed out a lot of the jet AI states as well, but those will come in a separate PR.

This PR aims to provide backwards compatibility with the existing aircraft code to an extent. I'll be addressing that functionality in the future as well as we move to the JetAIUpdate states.


tgjones (approved):
This is incredible work!

PR #992, opened 2024-09-06
Construct StateMachine States with GameObject, GameContext, and AIUpdate

charliefoxtwo:
mostly refactor work to build out all state with the required parameters. This should give us at least most of what we need within the states to start implementing them 🤞 

I also moved some of the specific states (Chinook, Hacker, Jet) to their own folders just to make it easier to sort through everything (I suspect this will really come into play as we add more states in the future).


charliefoxtwo:
@tgjones if you wanted to pass StateMachine instead of AIUpdate let me know and I can make those changes.

tgjones:
> if you wanted to pass StateMachine instead of AIUpdate let me know and I can make those changes

I think I would prefer that (and then add the relevant properties to `State` via `StateMachine`, e.g. `public GameObject GameObject => StateMachine.GameObject`)... but it's also fine to land this if you've already started building on top of it 😄

tgjones (approved):

charliefoxtwo:
I have started building on this, but it's not a big deal if we want to change directions a little bit. It's more important that the objects (GameObject, GameContext, AIUpdate) are accessible to me than _how_ they're accessible. If we want all StateMachines to have properties with those three items and then all States to take in a StateMachine, I'm totally fine with that too.

Do we want GameObject, GameContext, and AIUpdate to all be StateMachine properties?

tgjones:
> Do we want GameObject, GameContext, and AIUpdate to all be StateMachine properties?

I don't think we need `AIUpdate`, since it's reasonable to access it via `GameObject`, but the other two yes, I think so - and then for ergonomics, those properties could also be added to `State`. Something like

``` c#
public class StateMachineBase
{
    public GameObject GameObject { get; }
    public GameContext GameContext { get; }

    public StateMachineBase(GameObject gameObject, GameContext gameContext)
    {
        GameObject = gameObject;
        GameContext = gameContext;
    }
}

public abstract class State(StateMachineBase stateMachine)
{
    public GameObject GameObject => stateMachine.GameObject;
    public GameContext GameContext => stateMachine.GameContext;
    public AIUpdate AIUpdate => GameObject.AIUpdate;
}
```

charliefoxtwo:
Hm. We do still need to get the AIUpdateModuleData though. I'll play around with it.

charliefoxtwo:
@tgjones I went ahead and updated this so that States take in a StateMachine. The StateMachine is still constructed with the module data since that is needed for some of the states. Let me know if this is closer to what you had in mind!

charliefoxtwo:
As discussed on discord, will update to expose the relevant AIUpdate within the StateMachine

tgjones (approved):

PR #993, opened 2024-09-12
Rename UpdateFrame.Something to UpdateOrder

tgjones:
Small refactoring of `UpdateModule` based on parsing `UpdateFrame.Something` as probably meaning the execution order that update modules are updated. This value is always constant for a given module type. It is always either 0, 1, 2, or 3. Almost all module types are 2. Some examples:

- AIUpdate and its subclasses: 0
- PhysicsBehavior: 1
- _every other update module that I've seen_: 2
- ObjectWeaponStatusHelper: 3
- ObjectFiringTrackerHelper: 3

The order makes some sense - `AIUpdate` does the majority of stuff, including applying physics forces etc. Then `PhysicsBehavior` updates object's position and angle. Then all the other `UpdateModule`s do their thing (although this would imply that any physics forces applied by these update modules don't get applied till the next frame, which would be an interesting way to confirm this theory). Then `ObjectWeaponStatusHelper` and `ObjectFiringTrackerHelper` (which, guessing from the state stored in it, takes care of continuous fire weapons) which respond to changes in weapon state caused by everything before them.

My guess is that it only happens to be persisted because it's bit-packed into the `NextUpdateFrame` value. If it wasn't, they probably wouldn't have bothered persisting it. But they also didn't bother to remove it from the bit-packed value before persisting.

And the value can't be any higher than 3, because it's bit-packed into the low 2 bits of `NextUpdateFrame`. Because of the way it's bit-packed, you could do a simple sort of the uint value for `NextUpdateFrame` for all update modules, and that would give you the correct order of `UpdateModule`s in frame order, and then in execution order for that frame. Perhaps that's what the original engine did.

### DIFF
--- a/src/OpenSage.Game/Logic/Object/BehaviorModule.cs
+++ b/src/OpenSage.Game/Logic/Object/BehaviorModule.cs
@@ -7,8 +7,6 @@ namespace OpenSage.Logic.Object
 {
     public abstract class BehaviorModule : ModuleBase
     {
-        internal virtual void Update(BehaviorUpdateContext context) { }
-
         internal virtual void OnDie(BehaviorUpdateContext context, DeathType deathType, BitArray<ObjectStatus> status) { }
 
         internal virtual void OnDamageStateChanged(BehaviorUpdateContext context, BodyDamageType fromDamage, BodyDamageType toDamage) { }
@@ -52,7 +50,7 @@ namespace OpenSage.Logic.Object
         }
     }
 
-    public readonly struct LogicFrame
+    public readonly struct LogicFrame : IEquatable<LogicFrame>
     {
         public static readonly LogicFrame Zero = default;
         public static readonly LogicFrame MaxValue = new LogicFrame(uint.MaxValue);
@@ -99,10 +97,26 @@ namespace OpenSage.Logic.Object
             return left.Value >= right.Value;
         }
 
+        public static bool operator ==(LogicFrame left, LogicFrame right)
+        {
+            return left.Value == right.Value;
+        }
+
+        public static bool operator !=(LogicFrame left, LogicFrame right)
+        {
+            return left.Value != right.Value;
+        }
+
         public override string ToString()
         {
             return Value.ToString();
         }
+
+        public override bool Equals(object obj) => obj is LogicFrame frame && Equals(frame);
+
+        public bool Equals(LogicFrame other) => Value == other.Value;
+
+        public override int GetHashCode() => Value.GetHashCode();
     }
 
     public readonly struct LogicFrameSpan
--- a/src/OpenSage.Game/Logic/Object/Behaviors/BuildingBehavior.cs
+++ b/src/OpenSage.Game/Logic/Object/Behaviors/BuildingBehavior.cs
@@ -4,7 +4,7 @@ using OpenSage.Data.Ini;
 namespace OpenSage.Logic.Object
 {
     [AddedIn(SageGame.Bfme)]
-    public class BuildingBehavior : BehaviorModule
+    public class BuildingBehavior : UpdateModule
     {
         private readonly BuildingBehaviorModuleData _moduleData;
         private readonly GameObject _gameObject;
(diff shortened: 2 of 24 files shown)


src/OpenSage.Game/Logic/Object/Behaviors/BuildingBehavior.cs
```diff
@@ -4,7 +4,7 @@
 namespace OpenSage.Logic.Object
 {
     [AddedIn(SageGame.Bfme)]
-    public class BuildingBehavior : BehaviorModule
+    public class BuildingBehavior : UpdateModule
```
tgjones, line 7: From C&C3 XSDs, this is the correct base class.

src/OpenSage.Game/Logic/Object/Behaviors/CastleBehavior.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace OpenSage.Logic.Object
 {
-    internal sealed class CastleBehavior : BehaviorModule
+    internal sealed class CastleBehavior : FoundationAIUpdate
```
tgjones, line 12: From C&C3 XSDs (yes, this module is really there), this is the correct base class.

src/OpenSage.Game/Logic/Object/BehaviorModule.cs
```diff
@@ -52,7 +50,7 @@ public BehaviorUpdateContext(
         }
     }
 
-    public readonly struct LogicFrame
+    public readonly struct LogicFrame : IEquatable<LogicFrame>
```
charliefoxtwo, line 53: thank you

src/OpenSage.Game/Logic/Object/Behaviors/GateOpenAndCloseBehavior.cs
```diff
@@ -8,7 +8,7 @@
 namespace OpenSage.Logic.Object
 {
     [AddedIn(SageGame.Bfme)]
-    public class GateOpenAndCloseBehavior : BehaviorModule
+    public class GateOpenAndCloseBehavior : UpdateModule
```
tgjones, line 11: From C&C3 XSDs, this is the correct base class.

src/OpenSage.Game/Logic/Object/Behaviors/SpawnBehavior.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace OpenSage.Logic.Object
 {
-    internal sealed class SpawnBehavior : BehaviorModule
+    internal sealed class SpawnBehavior : BehaviorModule, IUpdateModule
```
tgjones, line 10: This module doesn't appear to inherit from `UpdateModule`, either based on C&C3 XSDs or its `Persist` method. But it does need to be updated, so I've added this `IUpdateModule` to allow non-`UpdateModule` classes to participate in updating, without involving _every_ `BehaviorModule` class.

src/OpenSage.Game/Logic/Object/Update/AutoFindHealingUpdate.cs
tgjones, line 1: This is based on what I saw in .sav files.

src/OpenSage.Game/Logic/Object/Update/AutoFindHealingUpdate.cs
```diff
@@ -4,7 +4,28 @@ namespace OpenSage.Logic.Object
...
     {
-        private uint _unknown;
+        private readonly AutoFindHealingUpdateModuleData _moduleData;
+
+        public AutoFindHealingUpdate(AutoFindHealingUpdateModuleData moduleData)
+        {
+            _moduleData = moduleData;
+        }
+
+        private LogicFrameSpan _framesTillNextScan;
```
charliefoxtwo, line 14: nit: `Until` or `Til` (I'd prefer the former personally) (to much of a grammar queen to not say anything lol)

charliefoxtwo (approved):

charliefoxtwo (approved):

PR #1000, opened 2024-10-03
Capture mouse scroll events over window controls

### LINKED ISSUE #698: Generals: Scrolling in skirmish menu zooms in/out of shellmap
It doesn't do this in the main menu, but it does work in the skirmish screen for some reason.
image
image

valeronm:
This adds mouse coordinates to mouse wheel events, so we can check if there is a window control under the mouse when scroll event occurs and handle that event on window level. This fixes ability to zoom map in main menu and also in-game when mouse is over controls like command bar and other windows. Fixes #698

### DIFF
--- a/src/OpenSage.Game/Logic/Object/Update/AIUpdate/AIUpdate.cs
+++ b/src/OpenSage.Game/Logic/Object/Update/AIUpdate/AIUpdate.cs
@@ -1,4 +1,4 @@
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Numerics;
 using OpenSage.Data.Ini;
 using OpenSage.Logic.AI;
@@ -100,7 +100,7 @@ namespace OpenSage.Logic.Object
 
             SetLocomotor(LocomotorSetType.Normal);
 
-            if (ModuleData.Turret != null)
+            if (ModuleData != null && ModuleData.Turret != null)
             {
                 _turretAIUpdate = ModuleData.Turret.CreateTurretAIUpdate(GameObject);
             }
--- a/src/OpenSage.Game/Logic/Object/Update/AIUpdate/AIUpdate.cs
+++ b/src/OpenSage.Game/Logic/Object/Update/AIUpdate/AIUpdate.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Numerics;
 using OpenSage.Data.Ini;
 using OpenSage.Logic.AI;
@@ -100,7 +100,7 @@ namespace OpenSage.Logic.Object
 
             SetLocomotor(LocomotorSetType.Normal);
 
-            if (ModuleData != null && ModuleData.Turret != null)
+            if (ModuleData.Turret != null)
             {
                 _turretAIUpdate = ModuleData.Turret.CreateTurretAIUpdate(GameObject);
             }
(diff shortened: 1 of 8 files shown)


charliefoxtwo (approved):
The changes look fine to me.

@tgjones do you know if there's a reason we use an explicit struct layout for `InputMessageValue`? I assume we're constructing a new one basically every frame, just curious whether reducing the size actually has a performance impact here, especially if we're able to just pass it as an `in` parameter everywhere now.

valeronm:
@charliefoxtwo Thanks for review. Looks like there is a problem with finding SDL2 library in Ubuntu and macOS on CI , so I just marked tests instantiating GameWindow as Windows-only for now.

src/OpenSage.Game/Logic/Object/Update/AIUpdate/AIUpdate.cs
```diff
@@ -98,7 +98,7 @@ internal AIUpdate(GameObject gameObject, GameContext context, AIUpdateModuleData
 
             SetLocomotor(LocomotorSetType.Normal);
 
-            if (ModuleData.Turret != null)
+            if (ModuleData != null && ModuleData.Turret != null)
```
tgjones, line 101: When did you find that `ModuleData` is `null`? I don't think that should ever be the case.

tgjones (approved):

tgjones:
> do you know if there's a reason we use an explicit struct layout for InputMessageValue?

I didn't measure performance either before or after making that struct use explicit layout. I suppose I just default to that when it's an either/or kind of struct.

src/OpenSage.Game/Logic/Object/Update/AIUpdate/AIUpdate.cs
charliefoxtwo: I have actually run into this same issue locally recently when using the sav2json tool. I never paid much attention to what caused it...

valeronm: It just raised `null` error for me locally all the time and I haven't checked why this happens. Most likely this isn't proper fix, so I removed this change from PR and may try investigate it deeper some later.

tgjones:
As it happens I also just landed a fix to make BitArray more thread safe, so you don’t need that particular change.

valeronm:
> As it happens I also just landed a fix to make BitArray more thread safe, so you don’t need that particular change.

Yeah, I made similar change but using `lock` but then found that it's already handled in master, so just rebased all the code discarding my change of BitArray.

src/OpenSage.Game/Logic/Object/Update/AIUpdate/AIUpdate.cs
valeronm: @tgjones It happens on loading ShellMap in Generals for me. `NullReferenceException` happens on execution of this [code](https://github.com/OpenSAGE/OpenSAGE/blob/490d996c320278833243773da6012adeb33cfc47/src/OpenSage.Game/Logic/Object/Update/AIUpdate/JetAIUpdate.cs#L420) in JetAIUpdate.cs. It triggers `AIUpdate` constructor and [this assignment](https://github.com/OpenSAGE/OpenSAGE/blob/490d996c320278833243773da6012adeb33cfc47/src/OpenSage.Game/Logic/Object/Update/AIUpdate/AIUpdate.cs#L94) doesn't work. `moduleData` is an instance of `JetAIUpdateModuleData` but `ModuleData` remains `null` after the assignment.

charliefoxtwo: Ah yup, that's it, and I have it fixed locally as part of the jet state stuff I'm working on.

PR #1001, opened 2024-10-05
Consolidate PersistBase

charliefoxtwo:
This is used in a lot of places, and consequently BeginObject and EndObject are forgotten in a lot of places. I got tired of forgetting to add it myself, and also got tired of typing `reader.BeginObject("Base")` - hopefully this makes it more difficult to forget? 🤷

### DIFF
--- a/src/OpenSage.Game/Logic/AI/AIStates/AIState32.cs
+++ b/src/OpenSage.Game/Logic/AI/AIStates/AIState32.cs
@@ -16,7 +16,7 @@ internal sealed class AIState32 : FollowWaypointsState
     {
         reader.PersistVersion(1);
 
-        base.Persist(reader);
+        reader.PersistBase(base.Persist);
 
         reader.PersistObject(_stateMachine);
     }
@@ -32,7 +32,7 @@ internal sealed class AIState32 : FollowWaypointsState
         {
             reader.PersistVersion(1);
 
-            base.Persist(reader);
+            reader.PersistBase(base.Persist);
         }
     }
 
--- a/src/OpenSage.Game/Logic/AI/AIStates/AIState6.cs
+++ b/src/OpenSage.Game/Logic/AI/AIStates/AIState6.cs
@@ -16,7 +16,7 @@ internal sealed class AIState6 : MoveTowardsState
     {
         reader.PersistVersion(1);
 
-        base.Persist(reader);
+        reader.PersistBase(base.Persist);
 
         reader.PersistInt32(ref _unknownInt);
         reader.PersistBoolean(ref _unknownBool1);
--- a/src/OpenSage.Game/Logic/AI/AIStates/AttackAreaState.cs
+++ b/src/OpenSage.Game/Logic/AI/AIStates/AttackAreaState.cs
@@ -41,9 +41,7 @@ namespace OpenSage.Logic.AI.AIStates
         {
             reader.PersistVersion(1);
 
-            reader.BeginObject("Base");
-            base.Persist(reader);
-            reader.EndObject();
+            reader.PersistBase(base.Persist);
         }
     }
 }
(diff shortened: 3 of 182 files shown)


tgjones (commented):
I really like the idea here - it makes the code look cleaner.

However, I'm concerned about all the extra memory allocations caused by the implicit `new Action(...)` every time you call `PersistBase`. I know we've got plenty of unnecessary memory allocations all over the place, and maybe it's fine to add more. But I did put some effort in to avoid unncessary memory allocation in state persistence code, which is why we use `static` lambdas in callbacks like this:

``` c#
reader.PersistArray(
    _disabledTypesFrames,
    static (StatePersister persister, ref LogicFrame item) =>
    {
        persister.PersistLogicFrameValue(ref item);
    });
```

Unfortunately static lambdas don't work when calling a `base` method. IIRC that's why I didn't use a lambda in the first place for persisting the base object.

Here's a [SharpLab snippet](https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzmAPgAQEwEYCwAUEZgMwAEO5AQqgKZEDeR5rlFAblAE4ICuEADaUALOQAKdbiigoEACgCULNs0JtyAXxWsd5AA7cA9gjoBjUwBNREqTLm0UdeZlzZyZoYOAQzAa2V1VT0NT0FvXz8lAG49bUJ4kgoqABEpKA46axAaeiY9QwyIU3JuOghLIxhBAE9KN3IAfU8zAAss2KDdLvZyNO4MrKU9NQ0NZt926wBechg6AHd67HkfJwA6SWlZBEVOjUSesnIjTO4ByzobLfsFQI1RsdYbncdnNbpNux29kLYAen+Lwc9HkEzaWV+PXimiAA==) showing the extra cost in assembly code. In `Derived.Persist`, you can toggle between the two `PersistBase` calls to see in the assembly for `Derived.Persist` how much it helps to cache the delegate - but then, you're losing the benefit of a nice callsite.

tgjones:
That said, if you still want to land it, I won't stand in the way 😄 (It must have taken a while to change all the callsites!)

charliefoxtwo:
Interesting note on memory, thanks! Given that we only call persist when saving and when loading (we don't call it every frame, right?) is resource usage that big of a deal here?

Not worried about the time spent preparing the pr - IDEs make things easy and it's fun tinkering with the code either way :)

tgjones:
> Given that we only call persist when saving and when loading (we don't call it every frame, right?) is resource usage that big of a deal here?

Probably not, at least in the current usage, although I did have the vague thought that this state persistence stuff could be reused for the checksum code that we’ll want later on, for detecting network mismatches. And that will be called during gameplay (but not every frame).

charliefoxtwo:
Ah, interesting thought. We probably would want to run it with some regularity still so that recovery from a mismatched state (assuming we'd offer such a thing) isn't so jarring.

The other option here is for me to just correct the existing call sites to properly use reader.BeginObject and EndObject. If all call sites are corrected then at least we reduce the risk of copypasta error in the future.

I am curious (as someone who doesn't normally pay too much attention to these things) - do you think we'll be working with a sufficient number of objects that the additional allocations will result in a performance impact? Alpine Assault has like 700 objects iirc. So maybe worst case we assume 10k (player units and structures don't usually number too high). Do you see this being an issue on modern hardware if we ran this even every logic (haven't had a chance to check out that SharpLab snippet yet)? If so, happy to update the call sites.

charliefoxtwo:
I'll also add it would be a single find and replace to update the call sites since they're all uniform right now, so basically no additional effort for me to switch this PR to explicit begin/end object statements everywhere.

tgjones:
Sorry for the rather belated reply :)

> do you think we'll be working with a sufficient number of objects that the additional allocations will result in a performance impact?

I do think so. I'm hoping we'll be able to avoid all GC-related pauses during normal gameplay, and one of the prerequisites for that is not allocating any memory during normal gameplay.

> The other option here is for me to just correct the existing call sites to properly use reader.BeginObject and EndObject. If all call sites are corrected then at least we reduce the risk of copypasta error in the future.

I think this is my preferred option - updating all call sites to properly call Begin / End. If you're still willing to do it, of course 😄

charliefoxtwo:
Yeah I'll have to take a look at all the merge conflicts, but I can take care of it

PR #1076, opened 2025-03-02
Fix incorrect virtual property access in constructor in AIUpdate.cs

paavohuhtala:
This caused Generals to instantly crash in the main menu, but it works after the change.


tgjones (approved):
Thanks!

PR #1294, opened 2025-03-03
Port ArmorUpgrade from Generals OSS release

### LINKED ISSUE #1247: Port GameLogic/Object/Upgrade/ArmorUpgrade.cpp

tgjones:
Resolves #1247.

* Extract `BitArrayMatchFinder` out of `W3dModelDraw` - this is used for armor sets as well as model condition states
* Store current armor in `ActiveBody`, not `GameObject`
* Add some upgrade and terrain decal stuff that got dragged in, even though that part of `ArmorUpgrade` seems like a huge hack (hopefully we can improve that later on)

### DIFF
--- a/src/OpenSage.Game/Logic/Object/Draw/DrawModule.cs
+++ b/src/OpenSage.Game/Logic/Object/Draw/DrawModule.cs
@@ -30,6 +30,11 @@ namespace OpenSage.Logic.Object
 
         public virtual void SetSupplyBoxesRemaining(float boxPercentage) { }
 
+        public virtual void SetTerrainDecal(ObjectDecalType decalType)
+        {
+            // TODO(Port): Implement this for subclasses.
+        }
+
         internal abstract void Update(in TimeInterval time);
 
         internal abstract void SetWorldMatrix(in Matrix4x4 worldMatrix);
--- a/src/OpenSage.Game/Utilities/BitArrayMatchFinder.cs
+++ b/src/OpenSage.Game/Utilities/BitArrayMatchFinder.cs
@@ -1,0 +1,67 @@
+﻿using System;
+using OpenSage.Logic.Object;
+using OpenSage.Mathematics;
+
+namespace OpenSage.Utilities;
+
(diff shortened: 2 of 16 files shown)


src/OpenSage.Game/Utilities/BitArrayMatchFinder.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
+using OpenSage.Logic.Object;
+using OpenSage.Mathematics;
+
+namespace OpenSage.Utilities;
+
+// TODO(Port): Check if the logic here matches SparseMatchFinder.h.
```
charliefoxtwo, line 7: just as a heads-up, I'm pretty sure I wrote most of this logic originally. I wrote a bunch of unit tests for ModelConditionState and it passed those but... things just never really looked right in game. I don't know that this code is actually correct, so it will be good to take a look when porting SparseMatchFinder.

I'm also glad you moved it, because now your name is in the commit history for it 😆

src/OpenSage.Game/Logic/Object/Draw/DrawModule.cs
```diff
@@ -30,6 +30,11 @@ public virtual void SetAnimationDuration(LogicFrameSpan frames) { }
 
         public virtual void SetSupplyBoxesRemaining(float boxPercentage) { }
 
+        public virtual void SetTerrainDecal(ObjectDecalType decalType)
+        {
+            // TODO(Port): Implement this for subclasses.
```
charliefoxtwo, line 35: Up to you, but personally I'd recommend creating a github issue for todos like this so they don't get lost.

charliefoxtwo (approved):

src/OpenSage.Game/Logic/Object/Draw/DrawModule.cs
tgjones: I actually prefer a TODO in the code (but that's just my personal preference 😄). I don't think there's a danger of it getting lost, it's easily searchable, and it's less overhead than creating an issue.

PR #1295, opened 2025-03-03
Port SidesList from Generals OSS release

### LINKED ISSUE #1039: Port GameLogic/Map/SidesList.cpp

paavohuhtala:
Resolves #1039.

- Update `BuildListInfo` based on Generals OSS release
- Update `SidesList` based on Generals OSS release
  - This includes a few iffy things, most notably `PrepareForMpOrSkirmish` and the related skirmish fields - I think it's essentially UI state that was stuck here for unknown reason.
  - `PrepareForMpOrSkirmish` is quite similar to `SideListUtility.CreateTeamsFromScbFile` - we got pretty close! I didn't remove our version yet or take the new method into use, because the logic is not yet complete without porting parts of `GameLogic.cpp`.
- I don't think either class belongs under `OpenSage.Data.Map` namespace anymore, because they are not (just) map data. `OpenSage.Logic.Map` to mirror `GameLogic/Logic/Map` in the C++ code?


src/OpenSage.Game/Data/Map/BuildListInfo.cs
```diff
@@ -0,0 +1,224 @@
+﻿using System.IO;
+using System.Numerics;
+using OpenSage.Data.Ini;
+using OpenSage.FileFormats;
+
+namespace OpenSage.Data.Map
+{
+    // TODO: Does this belong under Data.Map anymore?
```
tgjones, line 8: Yeah, let's move this to `OpenSage.Logic.Map` (but doesn't have to be in this PR). In general (ha), let's follow the C++ folder structure unless there's an actual reason to differ.

tgjones (approved):
💯

PR #1298, opened 2025-03-04
The big reformat

tgjones:
This PR applies the rules that were already (mostly) in our `.editorconfig` file.

I did add one new rule - to use file-scoped namespaces - and applied that.

I've also added a step to CI to verify there are no formatting violations.


charliefoxtwo:
image
Should we make sure this branch is up to date first so we don't have to worry about making additional commits for missing files?

tgjones:
> Should we make sure this branch is up to date first so we don't have to worry about making additional commits for missing files?

The only missing commit is the one you just merged, which modifies `src/Directory.Packages.props`, which isn't touched by this PR. So it should be fine.

charliefoxtwo (approved):
I checked through every file and this looks good to me

tgjones:
@charliefoxtwo thanks for reviewing!

PR #1299, opened 2025-03-05
Fixed nullability warnings.

vitimiti:
Introduce Debug.Assert statements to ensure critical objects are not null before proceeding with operations.

These checks improve code safety and debugging, while minimizing the impact on release builds.

The aim of this merge request is **not** to improve or add, but to ensure code maintainability as the project moves forwards with its larger goals.

### DIFF
--- a/src/OpenSage.Game/Logic/AI/AIStates/ChinookAIUpdate/ChinookCombatDropState.cs
+++ b/src/OpenSage.Game/Logic/AI/AIStates/ChinookAIUpdate/ChinookCombatDropState.cs
@@ -36,11 +36,13 @@ internal sealed class ChinookCombatDropState : State
     {
         reader.PersistVersion(2);
 
-        reader.PersistListWithUInt32Count(_ropes, (StatePersister persister, ref Rope? item) =>
+        #nullable disable // The callback is in a project with no nullability, but the Rope object is nullable.
+        reader.PersistListWithUInt32Count(_ropes, (StatePersister persister, ref Rope item) =>
         {
             item ??= new Rope();
             persister.PersistObjectValue(item);
         });
+        #nullable enable
     }
 
     private sealed class Rope : IPersistableObject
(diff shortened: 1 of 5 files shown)


charliefoxtwo:
Hey, thanks for putting this together!

I'm wondering if we can use this as an opportunity to spread null type checking further in the code base. Right now we're in this place where we're slowly enabling nullable reference types across the entire project. These build warnings aren't really harmful, but do serve as a reminder to us that this code isn't safe and needs to be made safe.

Is that something you'd be interested in tackling? It will probably be a bit more involved for each warning, but would provide great benefit! Otherwise I'm concerned by just asserting that the types aren't null, we might just be burying problems that could come back to bite us later.

vitimiti:
Hi, I am used to dealing with billable types, I'd be happy to modernize the code base to them, yes. I will do a different PR for it :D

vitimiti:
Work started here, for anybody that wishes to help or see progress: https://github.com/vitimiti/OpenSAGE/tree/modernize_base

charliefoxtwo:
Awesome! I see you have quite a few commits on that branch, so I'll just say that smaller PRs that only touch a few files at a time would be ideal where possible - makes them much easier to review :) (and also reduces the likelihood/severity of conflicts with other work going on by other contributors)

vitimiti:
That is a good point, I'll change to multiple PRs instead

charliefoxtwo:
I took a closer look at the changes you've started putting together, and I'd further recommend splitting into multiple PRs based on the purpose. It looks like you have a few different types of changes stacked up there:
- minor bug fixes (like the Matrix4x3 Equals method - nice catch)
- changes which make the code _safer_ to work with, as they impact the caller (nullable reference types)
- changes which make the code _easier_ to work with, as they don't impact the caller and only impact anyone editing the existing code (modernization)

Breaking these up into multiple PRs also helps us prioritize reviewing and get the important stuff (like bugfixes) merged as fast as possible while still allowing us to have thoughtful discussions about things like modernization with less pressure to merge the PR quickly to get other fixes in.

You've done a lot already, so nice job there!

vitimiti:
I will on my time off, I want to play this game natively on Linux :^)

PR #1302, opened 2025-03-07
Port PhysicsBehavior from Generals OSS release

### LINKED ISSUE #1194: Port GameLogic/Object/Update/PhysicsUpdate.cpp

tgjones:
Resolves #1194.

As expected, `PhysicsBehavior` uses a number of methods from elsewhere. I was as pragmatic as possible about implementing these. Where they were simple, I implemented them, and where they were not simple I stubbed them out. I _think_ in all the cases with a stub implementation, it doesn't actually matter because we don't have enough of the rest of the code yet to actually go into these codepaths.


src/OpenSage.Game/Logic/Object/Update/UpdateModule.cs
```diff
@@ -34,13 +41,28 @@ internal virtual void Update(BehaviorUpdateContext context)
 
         SetNextUpdateFrame(context.LogicFrame + FramesBetweenUpdates);
         RunUpdate(context);
+
+        var sleepTime = Update();
+        SetNextUpdateFrame(context.LogicFrame + sleepTime.FrameSpan);
+    }
+
+    // TODO: Remove other Update methods after everything uses this.
```
charliefoxtwo, line 49: I've seen this a couple places now. Should the "other" methods be marked as `Obsolete` perhaps?

src/OpenSage.Game/Logic/Object/Body/ActiveBody.cs
```diff
@@ -84,25 +89,28 @@ public override void DoDamage(DamageType damageType, Fix64 amount, DeathType dea
...
+        var takingDamage = damageInfo.Request.DamageType is not DamageType.Healing;
 
         var newHealth = takingDamage ? Health - actualDamage : Health + actualDamage;
 
         SetHealth(newHealth);
 
-        var damageRequest = new DamageDataRequest(damageDealer?.ID ?? 0, damageType, deathType, (float)amount, damageDealer?.Definition.Name ?? string.Empty);
-        var damageResult = new DamageDataResult((float)actualDamage, _lastHealthBeforeDamage - _currentHealth);
+        var damager = _context.GameLogic.GetObjectById(damageInfo.Request.DamageDealer);
+        damageInfo.Request.AttackerName = damager?.Definition.Name ?? string.Empty;
```
charliefoxtwo, line 102: Would it be better to set this on construction of the `DamageInfo` object? Or is this done here specifically to avoid setting this when the target's health is zero?

Just thinking if we don't need to set this here, this doesn't need to be a ref parameter and we don't need to expose the field to be written publicly when it _shouldn't_(?) change

src/OpenSage.Game/Logic/Object/Behaviors/PhysicsBehavior.cs
charliefoxtwo, line 1: I won't lie, there's no way I can reasonably review this lol

src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -138,6 +136,9 @@ internal set
         }
     }
 
+    // TODO: Remove this. There should be a better way to access the game context.
+    internal GameContext Context => _gameContext;
```
charliefoxtwo, line 140: This looks so familiar - I thought we actually removed this "recently"? What's using it now?

src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -535,6 +602,10 @@ void AddBehavior(string tag, BehaviorModule behavior)
                 {
                     _body = body;
                 }
+                else if (module is IContainModule contain)
```
charliefoxtwo, line 605: this statement is starting to seem like a good candidate for a switch expression

src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -1397,6 +1521,59 @@ private void CheckDisabledStates()
...
+    {
+        // TODO(Port): Implement this properly. See Object::clearDisabled()
+        UnDisable(type);
+    }
+
+    public void SetCaptured(bool isCaptured)
+    {
+        if (!isCaptured)
+        {
+            throw new ArgumentException("Clearing captured status. This should never happen.");
```
charliefoxtwo, line 1542: Is this true? What if a tech building is captured by a player, and then that player dies? Don't the objects revert to neutral?

(I'm assuming this behavior was ported, so I'm just curious)

charliefoxtwo (approved):

src/OpenSage.Game/Logic/Object/Body/ActiveBody.cs
tgjones: I was more-or-less copying the [original code](https://github.com/electronicarts/CnC_Generals_Zero_Hour/blob/main/GeneralsMD/Code/GameEngine/Source/GameLogic/Object/Body/ActiveBody.cpp#L370) here, although I'm sure it can be cleaned up a bit when we come to port `ActiveBody.cpp`.

I'm actually not sure yet whether it needs to be a `ref` parameter - I was copying what the original does, but I don't know if it's ever necessary for calling code to access the modified "request". I suggest we leave it as it is for now until we port this code properly.

src/OpenSage.Game/Logic/Object/Update/UpdateModule.cs
tgjones: I'm reluctant to do so just because it would add warnings. I'm expecting we'll gradually change everything to use the "correct" `Update` method as we port each `UpdateModule` across.

src/OpenSage.Game/Logic/Object/GameObject.cs
tgjones: I needed it [here](https://github.com/OpenSAGE/OpenSAGE/blob/377ffc35945bfdecebb0f485e7a5830b1bc02afd/src/OpenSage.Game/Logic/Object/Update/UpdateModule.cs#L63) and [here](https://github.com/OpenSAGE/OpenSAGE/blob/physicsbehavior/src/OpenSage.Game/Logic/Object/Behaviors/PhysicsBehavior.cs#L662). But you're right, I could propagate `GameContext` to the base `ObjectModule` class, so I'll do that.

src/OpenSage.Game/Logic/Object/GameObject.cs
tgjones: Yes, this was ported from [here](https://github.com/electronicarts/CnC_Generals_Zero_Hour/blob/0a05454d8574207440a5fb15241b98ad0b435590/GeneralsMD/Code/GameEngine/Source/GameLogic/Object/Object.cpp#L2001). I think there might be a few definitions of "captured", and this one here is specifically about the "private status" flag.

src/OpenSage.Game/Logic/Object/GameObject.cs
tgjones: Done (assuming you meant switch statement).

src/OpenSage.Game/Logic/Object/GameObject.cs
tgjones: Done.

PR #1305, opened 2025-03-08
Port camera controller from Generals

### LINKED ISSUE #490: Camera support rotation
In many rts game, you can hold some key like `Alt` which can rotate the camera, hope you guys can support this feature.

### LINKED ISSUE #677: Generals: Starting camera rotated 90 degrees clockwise
OpenSAGE:
image

Vanilla Generals:
image

paavohuhtala:
Resolves #677.
Resolves #490.

* Port camera controller from Generals / ZH. Most new code is in `W3DView` and its base class `View` to `TactivalView` (previously known as `RtsCameraController`), and from `LookAtXlat.cpp` / `LookAtTranslator` to `LookAtTranslator.`
  * Almost fully implemented gameplay camera system; biggest missing feature is view save / restore.
  * Starting camera is now at the right place!
  * Camera constraints are implemented and should be close to correct.
  * Works at arbitrary frame rates (tested at up to 240 FPS).
  * Most cinematic camera features are implemented. Still missing camera shake, and anything requiring post process shaders. Every method doesn't have a correspoding script action yet, but we can easily add those later.
* Expanded script action code generator to support nullable / optional int and float arguments (because some script actions changed with ZH).
* Many small changes I can't even remember at this point.

## Known issues

* Breaks 3D asset preview / arcball camera. We probably want to figure out its role and relationship to `Scene3D` before fixing it.
* Animation features are mostly untested aside from Generals shell map.


charliefoxtwo:
Does this by chance resolve https://github.com/OpenSAGE/OpenSAGE/issues/677?

paavohuhtala:
> Does this by chance resolve #677?

It does indeed! Adding that to description.

src/OpenSage.Core/DeltaTimer.cs
```diff
@@ -76,5 +76,13 @@ public TimeInterval(TimeSpan totalTime, TimeSpan deltaTime)
     public static TimeInterval Zero { get; } = new TimeInterval();
 
     // This is the ratio between stopwatch ticks and timespan ticks
+
     private static readonly double TickRatio = 10000000.0 / Stopwatch.Frequency;
+
+    private const double MsPerLogicFrame = 1000.0 / 30.0;
```
charliefoxtwo, line 82: I think this might already be hardcoded somewhere else - can we reuse that instead of hardcoding the logic frames per second in multiple places?

https://github.com/OpenSAGE/OpenSAGE/blob/a3cef36b59cd9b83065a6f4da8d6f19d1b29ff83/src/OpenSage.Game/Game.cs#L55

paavohuhtala: We should have this once in a central location, yes. We can't use `Game`'s constant in `OpenSage.Core` though, so maybe the constant should live somewhere in the core library.

charliefoxtwo: I think it does depend on the game though, since from the comment BFME's logic frame tick rate was 5hz rather than 30. Not sure of the best way to solve that.

tgjones: Why does this constant (and `LogicFrameRelativeDeltaTime `) need to be in `OpenSage.Core`? Could `LogicFrameRelativeDeltaTime ` instead be an extension method on `DeltaTimer` that's defined in `OpenSage.Game`?

src/OpenSage.Game/Game.cs
```diff
@@ -691,16 +685,24 @@ private void StartGame(
...
+                    Scene3D.TacticalView.SetZoomToDefault();
+                }
             }
         }
 
         if (Scene3D.LocalPlayer.SelectedUnits.Count > 0)
         {
             var mainUnit = Scene3D.LocalPlayer.SelectedUnits.First();
-            Scene3D.CameraController.GoToObject(mainUnit);
+            // Scene3D.CameraController.GoToObject(mainUnit);
```
tgjones, line 705: Can this be removed? (I haven't finished reviewing yet so I don't know there's code further down that performs the same behavior.)

src/OpenSage.Game/Graphics/Cameras/Camera.cs
```diff
@@ -1,14 +1,15 @@
...
+using OpenSage.Terrain;
 using Viewport = Veldrid.Viewport;
 
 namespace OpenSage.Graphics.Cameras;
 
 public sealed class Camera
 {
     private readonly Func<Viewport> _getViewport;
-    private Viewport _viewport;
+    public Viewport _viewport { get; private set; }
```
tgjones, line 12:
```suggestion
    public Viewport Viewport { get; private set; }
```

tgjones (approved):
This is great! 👍

src/OpenSage.Core/DeltaTimer.cs
paavohuhtala: Good idea! Shame that C# (still) doesn't support extension properties, but a method is fine too.

PR #1306, opened 2025-03-10
Port Locomotor from Generals

### LINKED ISSUE #1271: Port GameLogic/Object/Locomotor.cpp

tgjones:
Resolves #1271.

This is mostly a 1:1 port of `Locomotor.cpp`. It could use a refactoring pass, but I'll do that in a future PR.

The dozer now looks like this:

https://github.com/user-attachments/assets/d6b0316e-bff9-4518-80d5-57da85eb0c75


src/OpenSage.Mathematics/MathUtility.cs
```diff
@@ -43,18 +45,40 @@ public static float GetPitchFromDirection(in Vector3 direction)
...
+        return NormalizeAngle(alpha - beta);
+    }
+
+    public static float NormalizeAngle(float angle)
+    {
+        Debug.Assert(!float.IsNaN(angle), "Angle is NaN");
+
+        if (float.IsNaN(angle))
+        {
+            // Argh! Don't assert and then not handle it! Error bad! Fix error!
```
charliefoxtwo, line 67: Lol. Add a todo so we can find this later?

tgjones: This is actually a comment from the original code. I'm not sure if we should throw an exception here - I guess that's what the comment implies - but I think it's okay as-is, because we'll hit the assert in debug builds.

charliefoxtwo: Is this something that's worth logging? Just thinking about a case where we forget about this and then ship a build and them someone reports a weird bug, it might be nice if this at least surfaced in logs so we'd know this happened?

Amusing that this came from the original code lol

tgjones: For now I've thrown an exception - 7bf5891fbe3c4fc69d02f9c8e81361714a299baf

If we want to downgrade it to a logged message in the future, we can do that, but I think for now I'd rather know about these kind of issues in both debug and release builds.

charliefoxtwo (approved):

PR #1321, opened 2025-03-16
Steam version fixes

paavohuhtala:
- Fix `WndParser` crashing with ZH Steam version's `MainMenu.wnd`
  - The Steam version of base Generals doesn't have the same issue - looks like the Wnd file was modified by hand, and the ZH version has a small syntax issue which the game seems to handle without issues, but OpenSAGE didn't like it.
- Change .big file handling to override existing entries in case of duplicate filename
  - This is potentially a pretty big change, but it seems to match the original engine. I don't think `Patch` .big files could work any other way.
  - Without this change the actual load order was kind of inverted (vs how it is other games) - the files that were loaded first won, and there was no way to override their entries afterwards.
- Update .big file load order logic to match ZH Steam version
  - This is explained in comments in (hopefully) great detail 
- Support loading universal / language independent .str files
  - There is just one file like this (`Patch.str` from the current Steam version) from what I can tell.
- Load Generals / ZH Steam version's Patch.str if available
  - Required to display the `Custom Mission`  button properly in the single player menu

With these changes (with help from #1308), OpenSAGE can now boot into the shellmap in Zero Hour with the latest Steam build of the game.

### DIFF
--- a/src/OpenSage.Game/Content/ContentManager.cs
+++ b/src/OpenSage.Game/Content/ContentManager.cs
@@ -90,6 +90,15 @@ public sealed class ContentManager : DisposableBase
 
             switch (sageGame)
             {
+                // TOOD: We don't actually know how this file (which was added in the Steam Workshop update) gets loaded,
+                // as the publicly released source code is for a prior version.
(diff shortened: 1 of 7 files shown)


tgjones (approved):
Nice!

PR #1325, opened 2025-03-20
Port Body modules from Generals

### LINKED ISSUE #1077: Port GameLogic/Object/Body code

### LINKED ISSUE #1087: Port GameLogic/Object/Body/ActiveBody.cpp

### LINKED ISSUE #1088: Port GameLogic/Object/Body/BodyModule.cpp

### LINKED ISSUE #1089: Port GameLogic/Object/Body/HighlanderBody.cpp

### LINKED ISSUE #1090: Port GameLogic/Object/Body/ImmortalBody.cpp

### LINKED ISSUE #1091: Port GameLogic/Object/Body/InactiveBody.cpp

### LINKED ISSUE #1092: Port GameLogic/Object/Body/StructureBody.cpp

### LINKED ISSUE #1093: Port GameLogic/Object/Body/UndeadBody.cpp

tgjones:
Resolves #1077, resolves #1087, resolves #1088, resolves #1089, resolves #1090, resolves #1091, resolves #1092, resolves #1093

https://github.com/user-attachments/assets/83fb976e-f097-4351-b52f-0d96bb524563


src/OpenSage.Game/Graphics/Cameras/TacticalView.cs
```diff
@@ -607,7 +607,7 @@ internal void RotateCameraTowardPosition(Vector3 pos, double milliseconds, float
             angle = -angle;
         }
 
-        // Default camera is rotated 90 degrees, so match
+        // NoArmor camera is rotated 90 degrees, so match
```
charliefoxtwo, line 610: I won't comment all the places this happened, but looks like a refactor maybe went a bit awry?

tgjones: Thanks for spotting that. I let Visual Studio look in comments when renaming a property, but that was a bad idea in this case...

src/OpenSage.FileFormats/EnumUtility.cs
```diff
@@ -26,8 +27,8 @@ public static bool IsValueDefined<TEnum>(TEnum value)
     }
 
     public static int GetEnumCount<TEnum>()
-        where TEnum : Enum
+        where TEnum : struct, Enum
     {
-        return Enum.GetNames(typeof(TEnum)).Length;
+        return Enum.GetValuesAsUnderlyingType<TEnum>().Cast<int>().Max() + 1;
```
charliefoxtwo, line 32: What is this method supposed to do? When I read `GetEnumCount` I assume 'get the number of items in the enum", which the old implementation did and this doesn't. How is this intended to work for flags enums?

src/OpenSage.Game/Client/Drawable.cs
```diff
@@ -484,6 +484,31 @@ public bool GetFirstRenderObjInfo(
         return true;
     }
 
+    private static readonly Dictionary<BodyDamageType, ModelConditionFlag> DamageTypeLookup = new()
```
charliefoxtwo, line 487: Is it worth using `FrozenDictionary` since this collection is never updated? It takes a bit longer to instantiate, but especially since this is static readonly we'd only have to do it once and the performance is quite a bit better.

src/OpenSage.Game/Logic/Object/Locomotor.cs
```diff
@@ -1025,9 +1025,9 @@ private void MoveTowardsPositionWheels(GameObject obj, PhysicsBehavior physics,
             effectiveSlowDownDist = 1 * AIPathfind.PathfindCellSizeF;
         }
 
-        const float FIFTEEN_DEGREES = MathF.PI / 12.0f;
+        const float fifteenDegrees = MathF.PI / 12.0f;
         var projectFrames = _gameEngine.LogicFramesPerSecond / 2; // Project out 1/2 second.
-        if (MathF.Abs(relAngle) > FIFTEEN_DEGREES)
+        if (MathF.Abs(relAngle) > fifteenDegrees)
         {
             // If we're turning more than 10 degrees, check & see if we're moving into "impassable territory"
```
charliefoxtwo, line 1032: Assuming this is just ported over (and I realize you're just renaming this), but interesting that this is named fifteen degrees, but the comment says 10 degrees.

src/OpenSage.Game/Logic/Object/Update/BattlePlanUpdate.cs
```diff
@@ -43,12 +43,12 @@ public void ChangeBattlePlan(BattlePlanType newPlan)
...
-        ClearActiveBattlePlan();
-        // todo: remove unit bonuses (but don't disable units)
-        base.OnDie(context, deathType, status);
-    }
+    //internal override void OnDie(BehaviorUpdateContext context, DeathType deathType, BitArray<ObjectStatus> status)
+    //{
+    //    ClearActiveBattlePlan();
+    //    // todo: remove unit bonuses (but don't disable units)
+    //    base.OnDie(context, deathType, status);
+    //}
```
charliefoxtwo, line 51: guessing this is just to be implemented later?

src/OpenSage.FileFormats/EnumUtility.cs
tgjones: It gets the size of the array that is needed if you want to be able to use the enum values as integer indices into the array... which is subtely different from the number of items in the enum, if for example some values are repeated, or there's a non-contiguous range of values. I'll rename the method.

tgjones: > How is this intended to work for flags enums?

It doesn't work for flags enums.

charliefoxtwo: Ah, that makes a lot more sense for the behavior - yeah I think a rename would be good.

src/OpenSage.Game/Client/Drawable.cs
tgjones: Good idea, I've done that.

src/OpenSage.Game/Logic/Object/Locomotor.cs
tgjones: Ha, well spotted. I've updated the comment.

src/OpenSage.Game/Logic/Object/Update/BattlePlanUpdate.cs
tgjones: It is - I've added a comment explaining why it's commented out.

src/OpenSage.Game/Logic/Object/Body/UndeadBody.cs
```diff
@@ -1,7 +1,103 @@
...
+        // OnDie will cause the special behavior.
+        var total = 0;
+        foreach (var module in GameObject.FindBehaviors<SlowDeathBehavior>())
+        {
+            if (module.IsDieApplicable(damageInput))
+            {
+                total += module.GetProbabilityModifier(damageOutput);
+            }
+        }
+        Debug.Assert(total > 0, "Hmm, this is wrong");
```
charliefoxtwo, line 81: We're not wrapping this yet, are we?

src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -1047,7 +994,10 @@ internal void PrepareConstruction()
             return;
         }
 
-        Health = Fix64.Zero;
+        // TODO: This code should end up in DozerAIUpdate.Construct.
+        // Newly constructed objects start at one hit point.
+        BodyModule.InternalChangeHealth(-BodyModule.Health + 1.0f);
```
charliefoxtwo, line 999: I'm surprised we never saw this in sav files!

charliefoxtwo (approved):

src/OpenSage.Game/Logic/Object/Body/UndeadBody.cs
tgjones: Not yet - when I do, I'll search for all the existing calls.

charliefoxtwo: yup, that makes sense to me, just wanted to make sure we didn't implement that already :)

PR #1326, opened 2025-03-23
Split client and logic random values

tgjones:
I've ported `RandomValue.cpp` from the original game, but I'm not tied to it. We may want to switch back to `System.Random` as the underlying generator, if we're not aiming for bit-level compatibility with the original game.

The main change here is to have a clean split between client and logic random values.

### DIFF
--- a/src/OpenSage.Game/Logic/Object/GameObject.cs
+++ b/src/OpenSage.Game/Logic/Object/GameObject.cs
@@ -1476,7 +1476,7 @@ public sealed class GameObject : Entity, IInspectable, ICollidable, IPersistable
             .ToList();
 
         var sumProbabilityModifiers = slowDeathBehaviors.Sum(x => x.ProbabilityModifier);
-        var random = _gameEngine.GameLogic.Random.Next(0, sumProbabilityModifiers);
+        var random = _gameEngine.GameLogic.Random.Next(0, sumProbabilityModifiers - 1);
         var cumulative = 0;
         foreach (var deathBehavior in slowDeathBehaviors)
         {
--- a/src/OpenSage.Game/Logic/Object/GameObject.cs
+++ b/src/OpenSage.Game/Logic/Object/GameObject.cs
@@ -527,7 +527,7 @@ public sealed class GameObject : Entity, IInspectable, ICollidable, IPersistable
     {
         if (objectDefinition.BuildVariations != null && objectDefinition.BuildVariations.Count() > 0)
         {
-            objectDefinition = objectDefinition.BuildVariations[gameContext.Random.Next(0, objectDefinition.BuildVariations.Count())].Value;
+            objectDefinition = objectDefinition.BuildVariations[gameContext.GameLogic.Random.Next(0, objectDefinition.BuildVariations.Length - 1)].Value;
         }
 
         _objectMoved = true;
@@ -1476,7 +1476,7 @@ public sealed class GameObject : Entity, IInspectable, ICollidable, IPersistable
             .ToList();
 
         var sumProbabilityModifiers = slowDeathBehaviors.Sum(x => x.ProbabilityModifier);
-        var random = _gameEngine.Random.Next(sumProbabilityModifiers);
+        var random = _gameEngine.GameLogic.Random.Next(0, sumProbabilityModifiers);
         var cumulative = 0;
         foreach (var deathBehavior in slowDeathBehaviors)
         {
--- a/src/OpenSage.Game/Logic/Object/Update/AIUpdate/HackInternetAIUpdate.cs
+++ b/src/OpenSage.Game/Logic/Object/Update/AIUpdate/HackInternetAIUpdate.cs
@@ -105,7 +105,6 @@ internal sealed class HackInternetAIUpdateStateMachine : AIUpdateStateMachine
 
     internal LogicFrameSpan GetVariableFrames(LogicFrameSpan time, GameEngine gameEngine)
     {
-        // take a random float, *2 for 0 - 2, -1 for -1 - 1, *variance for our actual variance factor
         var variationFactor = AIUpdate.ModuleData.PackUnpackVariationFactor;
         var variation = gameEngine.GameLogic.Random.NextSingle(
             1.0f - variationFactor,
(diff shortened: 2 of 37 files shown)


penev92:
What does the original's random generator give you? 
Also was it actually random?

charliefoxtwo:
> Also was it actually random?
If this is the same random number generator that was used to select your army when you selected "Random", I seem to recall that was in fact not very random at all.

I guess if we went with our own RNG, it would mean loading an OG replay in opensage just wouldn't work. But at the same time, if the OG RNG isn't really that random (I don't know for sure whether it was or wasn't), then that seems like something we'd definitely want to change. I suppose we could also just use a different RNG if the user loads an old replay vs an opensage replay, and then use System.Random for everything that isn't an OG replay?

charliefoxtwo:
I wrote a little console script to test. It seems no less random than System.Random from my limited testing.

Performance, however, was way worse? I went ahead and using BenchmarkDotNet to just check the `Next` function.
```
BenchmarkDotNet v0.14.0, Windows 11 (10.0.26100.3476)
AMD Ryzen 9 5900X, 1 CPU, 24 logical and 12 physical cores
.NET SDK 9.0.104
  [Host]     : .NET 8.0.14 (8.0.1425.11118), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.14 (8.0.1425.11118), X64 RyuJIT AVX2
```
| Method       | Mean      | Error     | StdDev    |
|------------- |----------:|----------:|----------:|
| SystemRandom |  2.187 ns | 0.0043 ns | 0.0036 ns |
| GameRandom   | 11.650 ns | 0.1091 ns | 0.0967 ns |

src/OpenSage.Game/Logic/Object/Update/AIUpdate/HackInternetAIUpdate.cs
```diff
@@ -107,7 +106,11 @@ public HackInternetAIUpdateStateMachine(GameObject gameObject, GameEngine gameEn
     internal LogicFrameSpan GetVariableFrames(LogicFrameSpan time, GameEngine gameEngine)
     {
         // take a random float, *2 for 0 - 2, -1 for -1 - 1, *variance for our actual variance factor
```
charliefoxtwo, line 108: this comment probably isn't needed anymore now that you made the code readable lol

src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -1476,7 +1476,7 @@ private void ExecuteRandomSlowDeathBehavior(DeathType deathType)
             .ToList();
 
         var sumProbabilityModifiers = slowDeathBehaviors.Sum(x => x.ProbabilityModifier);
-        var random = _gameEngine.Random.Next(sumProbabilityModifiers);
+        var random = _gameEngine.GameLogic.Random.Next(0, sumProbabilityModifiers);
```
charliefoxtwo, line 1479: does this need to be `sumProbabilityModifiers - 1`? Not sure of the context here, figured I'd check to be safe

charliefoxtwo (approved):

penev92:
> that was in fact not very random at all

I figured as much. Old hand-made RNG and such...

> it would mean loading an OG replay in opensage just wouldn't work

Does OpenSAGE really need to be an OG replay viewer?

And with this being a game engine isn't performance a top priority?

tgjones:
I've refactored this. I've also set the default back to the `System.Random` implementation.

> Does OpenSAGE really need to be an OG replay viewer?

Not so much for users, but more for internal testing - it would be pretty nice if we could run a replay in OpenSAGE and the original game, and check that the state matches at every frame. It's a far-off goal, and maybe not realisticaly achievable, but I want to at least leave the door open for that possibility.

src/OpenSage.Mathematics/IRandom.cs
```diff
@@ -0,0 +1,152 @@
...
+    {
+        _random = CreateRandom((int)seed);
+    }
+
+    private static Random CreateRandom(int seed) => new(seed);
+
+    public int Next(int lo, int hi)
+    {
+        return _random.Next(lo, hi + 1);
+    }
```
charliefoxtwo, line 37: I think this is counterintuitive for most developers experienced with csharp. If I have a Random.Next method, I'm expecting the high value to be exclusive, not inclusive.

I know this is how the existing implementation functioned, so I think if we do something like this we should name the method something different (e.g. `NextInclusive` or similar) so it's clear what's happening here (I know that would result in many more lines changed).

Just my $0.02

charliefoxtwo (approved):

src/OpenSage.Mathematics/IRandom.cs
tgjones: That does make sense. On the other hand, I'd prefer for such a commonly-used method to have a short name. To be consistent, we'd have `NextSingleInclusive` or `NextLogicFrameSpanInclusive`, and that's arguably less readable. I'm not sure it's worth the benefit of clarity for developers used to `System.Random`. But I could be persuaded if you think we should.

charliefoxtwo: Yeah, makes sense. The name was just an example, but I don't know that I have a better one in mind.

If nothing else, I might at least suggest some clear documentation on the interface?

Continous:
> Does OpenSAGE really need to be an OG replay viewer?
> 
> And with this being a game engine isn't performance a top priority?

Even if we wanted a universal replay viewer it'd make sense to just support both RNG methods and just shelf the old as legacy code.

tgjones:
> Even if we wanted a universal replay viewer it'd make sense to just support both RNG methods and just shelf the old as legacy code.

I agree - that’s why I did that :)

src/OpenSage.Mathematics/IRandom.cs
tgjones: > If nothing else, I might at least suggest some clear documentation on the interface?

Good idea - I've [done that](https://github.com/OpenSAGE/OpenSAGE/pull/1326/commits/7f79f0dd2c60749c3805f44025fb2f191aaff623).

charliefoxtwo (approved):

PR #1328, opened 2025-03-25
Add ObjectId type

### LINKED ISSUE #209: Implement object IDs
Each `GameObject` should have a unique, sequential object id (32-bit). They should be assigned to map objects in processing order, and then assigned to new objects in a deterministic, monotonic way. 
They are present in the original SAGE, and can be seen in e.g replays.

Other things:

* The id counter should be stored in `Scene3D`. 
* `Scene3D` should support querying objects by their `ObjectId`.
* I've added "pseudo-object IDs" in #189, but since it's based on array indices they are not suitable for long-term use.

paavohuhtala:
Resolves #209 (which was actually done years ago)

- Add ObjectId struct
- Use ObjectId type for object IDs
- Rename GameObject.ID -> Id


src/OpenSage.Game/Logic/GameLogic.cs
```diff
@@ -85,28 +85,28 @@ public GameObject CreateObject(ObjectDefinition objectDefinition, Player player)
...
         }
-        _objects[(int)objectId] = gameObject;
+        _objects[(int)objectId.Id] = gameObject;
     }
 
-    public GameObject GetObjectById(uint id)
+    public GameObject GetObjectById(ObjectId id)
     {
-        return _objects[(int)id];
+        return _objects[(int)id.Id];
```
charliefoxtwo, line 109: Is it worth calling this like RawValue or something so we don't have `id.id` calls which might be confusing?

src/OpenSage.Game/Logic/Object/Behaviors/ParkingPlaceBehavior.cs
```diff
@@ -204,15 +204,15 @@ public void ReserveRunway(uint objectId)
         // if all runways are occupied, check if we can double-stack
         for (var i = 0; i < _runwayAssignments.Length; i++)
         {
-            if (_runwayAssignments[i].Aircraft2Id == 0)
+            if (!_runwayAssignments[i].Aircraft2Id.IsValid)
```
charliefoxtwo, line 207: Is it worth adding a convenience `Invalid` (or IsInvalid) property instead of all the negate calls? Seems like we're negating this a lot.

src/OpenSage.Game/Logic/Object/GameObject.cs
```diff
@@ -2031,7 +2031,7 @@ internal ObjectVeterancyHelper(GameObject gameObject, IGameObjectCollection game
 
     public bool GainExperience(int experience)
     {
-        if (_experienceSinkObjectId > 0)
+        if (_experienceSinkObjectId != ObjectId.Invalid)
```
charliefoxtwo, line 2034: Can this just be IsValid? I think there might be other places where this could be changed as well.

src/OpenSage.Game/Data/Rep/ReplayChunk.cs
```diff
@@ -54,7 +54,7 @@ internal static ReplayChunk Parse(BinaryReader reader)
                         break;
 
                     case OrderArgumentType.ObjectId:
-                        order.AddObjectIdArgument(reader.ReadUInt32());
+                        order.AddObjectIdArgument(new ObjectId(reader.ReadUInt32()));
```
charliefoxtwo, line 57: Is there a ReadObjectId method?

charliefoxtwo (approved):
Looks good, nice work! It seems like there are a few comparisons we could simplify (though they wouldn't necessarily need to be in this PR)

src/OpenSage.Game/Logic/Object/GameObject.cs
paavohuhtala: It definitely can! I think I added this check before I added `IsValid`. Find & replaced all of the simple cases I could find.

src/OpenSage.Game/Data/Rep/ReplayChunk.cs
paavohuhtala: Maybe there should be, but it's not easy to add as `BinaryReaderExtensions` live in `OpenSage.FileFormats` which doesn't (and cannot) reference `ObjectId`. We'd either need to add a separate set of binary reader extensions to this assembly, or to move `ObjectId` to `OpenSage.Core` and make `FileFormats` depend on that.

src/OpenSage.Game/Logic/Object/Behaviors/ParkingPlaceBehavior.cs
paavohuhtala: Added, and also added some documentation comments for ObjectId's fields.

src/OpenSage.Game/Logic/GameLogic.cs
paavohuhtala: Chnaging the name is a good idea. I'm not 100% sold on the name `RawValue` as that kind of implies there exists some sort of more refined or "safer" version of the value. For now I changed it to `Index`, as that what an object ID ultimately is (but 99.5% of code doesn't need to know that).

src/OpenSage.Game/Data/Rep/ReplayChunk.cs
tgjones: I'd suggest adding a separate set of binary reader extensions to this assembly (or... keep it as it is, since it's not needed in many places).

tgjones (approved):
Looks good!

src/OpenSage.Game/Data/Rep/ReplayChunk.cs
paavohuhtala: Looks like this is the only place for now where this exact pattern repeats (read uint32 + `new ObjectId`), so we probably don't need that extension for now.

PR #1329, opened 2025-03-28
Port ExperienceTracker & related code

### LINKED ISSUE #1268: Port GameLogic/Object/ExperienceTracker.cpp

paavohuhtala:
Resolves #1268.

Replaces `VeterancyHelper` with `ExperienceTracker`, ported from the [original C++ class](https://github.com/electronicarts/CnC_Generals_Zero_Hour/blob/main/GeneralsMD/Code/GameEngine/Source/GameLogic/Object/ExperienceTracker.cpp). Also includes some other related code I had to port, such as player / team relationship logic.

### DIFF
--- a/src/OpenSage.Game/Logic/Object/ExperienceTracker.cs
+++ b/src/OpenSage.Game/Logic/Object/ExperienceTracker.cs
@@ -1,0 +1,257 @@
+﻿#nullable enable
+
+namespace OpenSage.Logic.Object;
+
+public class ExperienceTracker(GameObject parent, GameEngine engine) : IPersistableObject
+{
+    private readonly GameEngine _engine = engine;
+
+    /// <summary>
+    /// Object I am owned by.
+    /// </summary>
+    private readonly GameObject _parent = parent;
+
+    /// <summary>
+    /// Level of experience.
+    /// </summary>
+    private VeterancyLevel _currentLevel = VeterancyLevel.Regular;
+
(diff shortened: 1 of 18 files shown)


src/OpenSage.Game/Logic/Object/ExperienceTracker.cs
```diff
@@ -0,0 +1,257 @@
...
+            levelIndex++;
+        }
+
+        _currentLevel = (VeterancyLevel)levelIndex;
+
+        if (oldLevel != _currentLevel)
+        {
+            OnVeterancyLevelChanged(oldLevel, _currentLevel, provideFeedback);
+            // Comment from original code: "paradox! this may be a level lost!"
+            // What does that mean?
```
charliefoxtwo, line 244: potentially this could be used to set a unit's experience to a lower amount than they currently have?

charliefoxtwo (approved):

src/OpenSage.Game/Logic/Object/ExperienceTracker.cs
paavohuhtala: True, though the same applies to `SetVeterancyLevel` as well.

charliefoxtwo: I don't think this needs to be solved now or anything, but it's probably worth revisiting at some point. I'm assuming mission scripts can set the veterancy level of a unit, even if it means lowering its rank.

PR #1337, opened 2025-03-29
Port locomotor-related code in Drawable

tgjones:
https://github.com/user-attachments/assets/9cc38342-eef7-40f1-a8dc-68f0ff5b6c44

TODO:

* [ ] Refactor the code. There's a lot of duplication between the different locomotor types. I can understand how that came to be, but I'd like to see how much of the duplicated code can be shared.
* [ ] Make this code handle deltaTime correctly. Right now it assumes a fixed 30fps

*Edit*: Actually I'll do the above two things in separate PRs. I'm having trouble understanding how to apply deltaTime to this drawable locomotor stuff, so I first want to refactor some of the client / logic timing code.

### DIFF
--- a/src/OpenSage.Game/Client/Drawable.cs
+++ b/src/OpenSage.Game/Client/Drawable.cs
@@ -361,6 +361,9 @@ public sealed class Drawable : Entity, IPersistableObject
                 break;
         }
 
+        var finalMatrix = InstanceMatrix * worldMatrix;
+        ApplyPhysicsTransform(ref finalMatrix);
+
         // Update all draw modules
         foreach (var drawModule in DrawModules)
         {
@@ -380,8 +383,6 @@ public sealed class Drawable : Entity, IPersistableObject
 
             drawModule.Update(gameTime);
 
-            var finalMatrix = InstanceMatrix * worldMatrix;
-            ApplyPhysicsTransform(ref finalMatrix);
             drawModule.SetWorldMatrix(finalMatrix);
 
             drawModule.BuildRenderList(
--- a/src/OpenSage.Game/Client/Drawable.cs
+++ b/src/OpenSage.Game/Client/Drawable.cs
@@ -1811,6 +1811,7 @@ public sealed class Drawable : Entity, IPersistableObject
 
             if (customized)
             {
+                // TODO(Port): Implement this.
                 throw new NotImplementedException();
             }
         }
@@ -2148,7 +2149,10 @@ public class DrawableInfo
         PotentialOccludee = 0x4,
         IsTranslucent = 0x8,
         IsNonOccluderOrOccludee = 0x10,
-        DelayedRender = IsTranslucent | PotentialOccludee,
+
+        // TODO(Port):
+        // Original has this enum value, but we'll instead do it with an extension method.
+        //DelayedRender = IsTranslucent | PotentialOccludee,
     }
 
     /// <summary>
--- a/src/OpenSage.Game/Client/Drawable.cs
+++ b/src/OpenSage.Game/Client/Drawable.cs
@@ -496,22 +496,19 @@ public sealed class Drawable : Entity, IPersistableObject
     {
         EnsureDrawableLocomotorInfo();
 
-        var ACCEL_PITCH_LIMIT = locomotor.LocomotorTemplate.AccelerationPitchLimit;
-        var DECEL_PITCH_LIMIT = locomotor.LocomotorTemplate.DecelerationPitchLimit;
-        var BOUNCE_ANGLE_KICK = locomotor.LocomotorTemplate.BounceAmount;
-        var PITCH_STIFFNESS = locomotor.LocomotorTemplate.PitchStiffness;
-        var ROLL_STIFFNESS = locomotor.LocomotorTemplate.RollStiffness;
-        var PITCH_DAMPING = locomotor.LocomotorTemplate.PitchDamping;
-        var ROLL_DAMPING = locomotor.LocomotorTemplate.RollDamping;
-        var FORWARD_ACCEL_COEFF = locomotor.LocomotorTemplate.ForwardAccelerationPitchFactor;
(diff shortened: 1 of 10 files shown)


src/OpenSage.Game/Logic/Object/Animation.cs
```diff
@@ -178,18 +178,26 @@ private enum AnimationDirection : ushort
...
+
     MoneyPickUp,
     LevelGainedAnimation,
     GetHealedAnimation,
-    BombTimed,
-    BombRemote,
-    CarBomb,
-    Disabled,
     AmmoFull,
     AmmoEmpty,
```
charliefoxtwo, line 202: out of curiosity, why do some of these not have values?

src/OpenSage.Game/Logic/Object/Update/AIUpdate/AIUpdate.cs
```diff
@@ -93,6 +93,20 @@ public class AIUpdate : UpdateModule
...
+    public float CurrentLocomotorSpeed
+    {
+        get
+        {
+            if (CurrentLocomotor != null)
+            {
+                return CurrentLocomotor.GetMaxSpeedForCondition(GameObject.BodyModule.DamageState);
+            }
+
+            Logger.Info("No current locomotor!");
```
charliefoxtwo, line 105: warn/error? Or is this fine? I guess I assume because it has an exclamation mark that it's not fine lol

src/OpenSage.Game/Client/Drawable.cs
```diff
@@ -363,6 +395,1110 @@ internal void BuildRenderList(RenderList renderList, Camera camera, in TimeInter
...
+                    _locomotorInfo.WheelInfo.RearRightHeightOffset = _locomotorInfo.WheelInfo.RearLeftHeightOffset;
+                }
+            }
+            // Calculate suspension info.
+            var length2 = obj.Geometry.MajorRadius;
+            //Real width = obj->getGeometryInfo().getMinorRadius();
+            var pitchHeight2 = length2 * MathF.Sin(_locomotorInfo.Pitch + _locomotorInfo.AccelerationPitch - groundPitch);
+            //Real rollHeight = width*Sin(_locomotorInfo.Roll + _locomotorInfo.AccelerationRoll - groundRoll);
+            info.TotalZ = MathF.Abs(pitchHeight2) / 4;// + MathF.Abs(rollHeight)/4;
+            //return; // maintain the same orientation while we fly through the air.
```
charliefoxtwo, line 886: do we not want to return here?

src/OpenSage.Game/Client/Drawable.cs
```diff
@@ -674,23 +1797,22 @@ public void Persist(StatePersister reader)
...
-            var unknownBool4 = true;
-            reader.PersistBoolean(ref unknownBool4);
-            if (!unknownBool4)
+            var customized = false;
+            reader.PersistBoolean(ref customized);
+
+            if (customized)
             {
-                throw new InvalidStateException();
+                throw new NotImplementedException();
```
charliefoxtwo, line 1814: does this need a todo port?

src/OpenSage.Game/Client/Drawable.cs
```diff
@@ -767,3 +2098,77 @@ public enum ObjectDecalType
...
+    [Flags]
+    public enum ExtraRenderFlags
+    {
+        None = 0,
+        IsOccluded = 0x1,
+        PotentialOccluder = 0x2,
+        PotentialOccludee = 0x4,
+        IsTranslucent = 0x8,
+        IsNonOccluderOrOccludee = 0x10,
+        DelayedRender = IsTranslucent | PotentialOccludee,
```
charliefoxtwo, line 2151: I'm guessing the way this will end up being used is like `if (ExtraRenderFlags.DelayedRender.HasFlag(variable))` since I don't think it works the other way around, which I could see being confusing.

I guess what I'm saying is we might have to be careful about how this is used. If nothing is ever _set_ to `DelayedRender` and it would only be used to check if something is of this type (and assuming if it has _either_ value it should be delayed), maybe an extension method would be better?

Or maybe that's not how it will be used at all 🤷

src/OpenSage.Game/Logic/Object/Animation.cs
tgjones: The ones with values are the ones that appear [here](https://github.com/electronicarts/CnC_Generals_Zero_Hour/blob/0a05454d8574207440a5fb15241b98ad0b435590/GeneralsMD/Code/GameEngine/Include/GameClient/Drawable.h#L71-L107). I don't know this area well enough to know why some of these, like `MoneyPickup`, are actually used in Zero Hour but don't appear in this enum.

src/OpenSage.Game/Logic/Object/Update/AIUpdate/AIUpdate.cs
tgjones: I actually copied the original in logging rather than asserting here. I don't know how carefully the original code decided between logging vs asserting, but for now I'm mostly just copying it.

Talking of which, I'd like to change the logging system so each class doesn't need its own logger instance. Building on #1343, maybe we should have a `DebugUtility.Log`.

src/OpenSage.Game/Client/Drawable.cs
tgjones: Correct, for motorcycles, we don't return here. Which means that motorcycles, unlike four-wheeled vehicles, change their pitch and roll while in the air. Presumably that's intentional.

src/OpenSage.Game/Client/Drawable.cs
tgjones: I agree, it would be nicer with an extension method. cf776357ec7d2ca4572d63f9be9dcdeab330d946

src/OpenSage.Game/Client/Drawable.cs
tgjones: Good catch - cf776357ec7d2ca4572d63f9be9dcdeab330d946

charliefoxtwo (approved):

PR #1338, opened 2025-03-29
Port Money.cpp / BankAccount and (some of) AcademyStats.cpp

paavohuhtala:
- Update `BankAccount` with ported code from Money.cpp
  - `BankAccount` is the name of our original implementation. I tried renaming it `Money` to match Generals, but since it has a field named `Money` (both in our version and in Generals), the compiler didn't allow it. I think our current name is better anyway.
- Add partial port of `AcademyStats` (which is used to generate end-of-game strategy tips that are shown in ZH's score screen)
  - This is a pretty long class and IMHO not very important in the grand scheme of things. I added all of the fields as it was a pretty simple mechanical process, and the method used by `BankAccount`.
  - I also implemented `Persist`. After finishing the implementation I noticed that ZH doesn't seem to ever call the `xfer` method, so even though there's a serialisation format it's probably not actually used anywhere. I decided to leave it in anyway, in case I was mistaken and it is actually called somewhere, or if we want to introduce an OpenSAGE-specific version of the Player chunk which includes this missing data.

### DIFF
--- a/src/OpenSage.Game/Logic/AcademyStats.cs
+++ b/src/OpenSage.Game/Logic/AcademyStats.cs
@@ -1,0 +1,224 @@
+﻿using OpenSage.Logic.Object;
+
+namespace OpenSage.Logic;
+
+// TODO(Port): Add stats collection methods as needed.
+public sealed class AcademyStats(IGame game) : IPersistableObject
+{
+    private readonly IGame _game = game;
+
+    private uint _nextUpdateFrame;
+    private bool _firstUpdate;
+    private bool _unknownSide;
+
+    // Tier 1 (Basic advice)
+
+    // Did player run out of money before building a supply center?
+    private bool _spentCashBeforeBuildingSupplyCenter;
+    private uint _supplyCentersBuilt;
+    private ObjectDefinition _supplyCenterTemplate;
+    uint _supplyCenterCost;
+
+    // Did player build radar (if applicable)?
+    private bool _researchedRadar;
+
+    // Did player build any dozers/workers?
+    private uint _peonsBuilt;
+
+    // Did player ever capture a structure?
+    private uint _structuresCaptured;
+
+    // Did player spend any generals points?
+    private uint _generalsPointsSpent;
+
+    // Did player ever use a generals power or superweapon?
+    private uint _specialPowersUsed;
+
+    // Did player garrison any structures?
+    private uint _structuresGarrisoned;
+
+    // How idle was the player in building military units?
+    private uint _idleBuildingUnitsMaxFrames;
+    private uint _lastUnitBuiltFrame;
+
+    // Did player drag select units?
+    private uint _dragSelectUnits;
+
+    // Did player upgrade anything?
+    private uint _upgradesPurchased;
+
+    // Was player out of power for more than 10 minutes?
+    private uint _powerOutMaxFrames;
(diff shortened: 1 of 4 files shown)


src/OpenSage.Game/Logic/Player.cs
```diff
@@ -83,7 +83,7 @@ public class Player : IPersistableObject
 
     public AIPlayer? AIPlayer { get; private set; }
 
-    public AcademyStats AcademyStats { get; } = new AcademyStats();
+    public readonly AcademyStats AcademyStats;
```
charliefoxtwo, line 86: Was there a reason this had to be changed from a property to a field? Is that just in case we choose to persist it in the future?

charliefoxtwo (approved):

src/OpenSage.Game/Logic/Player.cs
paavohuhtala: I used a couple of days old `master` as the base, and in that version there was no `AcademyStats` field / class yet, so I added it as a field. Tim had added a stub implementation in the mean time, so there was a conflict when rebasing. I solved it by just taking my changes (since the initialisation had to be moved to the constructor anyway because of `IGame` parameter). I can change it back to property; no strong feelings either way 😄

src/OpenSage.Game/Logic/AcademyStats.cs
```diff
@@ -0,0 +1,224 @@
+﻿using OpenSage.Logic.Object;
+
+namespace OpenSage.Logic;
+
+// TODO(Port): Add stats collection methods as needed.
+public sealed class AcademyStats(IGame game) : IPersistableObject
+{
+    private readonly IGame _game = game;
+
+    private uint _nextUpdateFrame;
```
tgjones, line 10:
```suggestion
    private LogicFrame _nextUpdateFrame;
```

src/OpenSage.Game/Logic/AcademyStats.cs
```diff
@@ -0,0 +1,224 @@
...
+    private bool _firstUpdate;
+    private bool _unknownSide;
+
+    // Tier 1 (Basic advice)
+
+    // Did player run out of money before building a supply center?
+    private bool _spentCashBeforeBuildingSupplyCenter;
+    private uint _supplyCentersBuilt;
+    private ObjectDefinition _supplyCenterTemplate;
+    uint _supplyCenterCost;
```
tgjones, line 20:
```suggestion
    private uint _supplyCenterCost;
```

src/OpenSage.Game/Logic/AcademyStats.cs
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    // TOOD: After porting this I realised that this doesn't seem to be used anywhere in ZH?
+    // Which means your stats will be reset every time you save & load a game.
+    // We could add this to the savefile, but we'd have to increment the version of the Player xfer format.
+    public void Persist(StatePersister persister)
+    {
+        persister.PersistVersion(1);
+
+        persister.PersistUInt32(ref _nextUpdateFrame);
```
tgjones, line 145:
```suggestion
        persister.PersistLogicFrame(ref _nextUpdateFrame);
```

tgjones (approved):

PR #1341, opened 2025-04-01
Port field names for AIUpdate

tgjones:


src/OpenSage.Game/Scripting/ScriptingSystem.cs
```diff
@@ -235,6 +235,19 @@ private ref ScriptingCounter GetCounter(string name)
         return ref result;
     }
 
+    public AttackPriority GetAttackInfo(string name)
+    {
+        for (var i = 0; i < _attackPriorities.Count; i++)
+        {
+            if (_attackPriorities[i].Name == name)
```
charliefoxtwo, line 242: Is this worth building and caching a dictionary?

charliefoxtwo (approved):

src/OpenSage.Game/Scripting/ScriptingSystem.cs
tgjones: That's a good idea, although the original game didn't, which means we still need to store / persist it as a list. Definitely something to consider when we properly port `ScriptingSystem`.

charliefoxtwo: I guess I'm missing a lot of context here, so I'm not sure how often this is called or whether `_attackPriorities` ever changes, but I was thinking basically we just cache something in memory, so the first time this method is called just build the dictionary and then use it for all future calls.

If the list is small, it's probably not a huge deal either way.

PR #1342, opened 2025-04-04
chore: implement crc calculation for maps

DarkAtra:
Based on https://github.com/electronicarts/CnC_Generals_Zero_Hour/blob/0a05454d8574207440a5fb15241b98ad0b435590/Generals/Code/GameEngine/Source/Common/crc.cpp#L55


tgjones (approved):
Nice, thank you! At some point we'll probably want this in a reusable `Crc` class but that can be done in a future PR.

PR #1343, opened 2025-04-04
Add `Debug.Assert` / `Debug.Fail` wrapper

tgjones:


src/OpenSage.Game/Logic/Object/Body/BodyModule.cs
```diff
@@ -122,13 +121,13 @@ public virtual void ClearLastAttacker() { }
     public virtual bool FrontCrushed
     {
         get => false;
-        set => Debug.Fail("You should never call this for generic Bodys");
+        set => DebugUtility.Crash("You should never call this for generic Bodys");
```
charliefoxtwo, line 124: Is this reachable, or should this be abstract/have an abstract setter? Not really related to this PR, but it feels like a code smell (same with the below property)

charliefoxtwo (approved):

src/OpenSage.Game/Logic/Object/Body/BodyModule.cs
tgjones: It is a bit strange. The getter is non-abstract, and returns the default value of `false`. The setter should really be abstract, but C#/.NET doesn't allow a non-abstract getter and an abstract setter.

charliefoxtwo: Yeah, by abstract setter I was more thinking an explicit abstract set method, since like you mentioned you can't have just an abstract setter on a property.

PR #1348, opened 2025-04-12
Fix build warnings

tgjones:


src/OpenSage.Game/Logic/Orders/OrderProcessor.cs
```diff
@@ -399,45 +408,50 @@ public void Process(IEnumerable<Order> orders)
...
+                            var behavior = unit.FindBehavior<SupplyAIUpdate>();
 
-                        if (supplyPoint.IsKindOf(ObjectKinds.SupplySource))
-                        {
-                            behavior.CurrentSupplySource = supplyPoint;
-                            behavior.SupplyGatherState = SupplyAIUpdate.SupplyGatherStates.SearchingForSupplySource;
-                        }
-                        else // if it's not a supply source, it's a supply center
-                        {
-                            behavior.CurrentSupplyTarget = supplyPoint;
```
charliefoxtwo, line 439: Would we not want to set this to null if for example the vehicles supply center was destroyed?

charliefoxtwo (approved):
Nice! Just the one question about the supply stuff

src/OpenSage.Game/Logic/Orders/OrderProcessor.cs
tgjones: Good point. I've had a go at improving it; although I'm not sure if I should really reset both. I imagine this will be fixed when we actually port this code.

PR #1351, opened 2025-04-20
Port some state machine code

tgjones:
This PR does some foundational refactoring of our state machine code, in preparation for porting some actual state classes. We were fairly close with our existing implementation. I've changed how states are configured to match the original, including specifying a "next" state, for both "success" and "failure" conditions.

I went through a few iterations on this, to see if there was a more elegant way to write this code. But because state classes (in particular the idle state) can be used in multiple different state machines, potentially with different IDs, it is necessary to have state IDs defined completely separately from the state classes they're associated with. So a direct port seemed like the best option.

### DIFF
--- a/src/OpenSage.Game/Logic/Object/Update/AIUpdate/JetAIUpdate.cs
+++ b/src/OpenSage.Game/Logic/Object/Update/AIUpdate/JetAIUpdate.cs
@@ -338,8 +338,13 @@ internal sealed class JetAIUpdateStateMachine : AIUpdateStateMachine
     {
         AIUpdate = aiUpdate;
 
-        AddState(WaitForAirfieldWinchesterState.StateId, new WaitForAirfieldWinchesterState(this));
+        DefineState(JetAIStateIds.CirclingDeadAirfield, new WaitForAirfieldWinchesterState(this), AIStateIds.Idle, AIStateIds.Idle);
     }
+}
+
+internal static class JetAIStateIds
+{
+    public static readonly StateId CirclingDeadAirfield = new(1013);
 }
 
 /// <summary>
--- a/src/OpenSage.Game/Logic/AI/AIStates/AIState32.cs
+++ b/src/OpenSage.Game/Logic/AI/AIStates/AIState32.cs
@@ -25,7 +25,11 @@ internal sealed class AIState32 : FollowWaypointsState
     {
         public AIState32StateMachine(AIUpdateStateMachine parentStateMachine) : base(parentStateMachine)
         {
-            AddState(IdleState.StateId, new IdleState(this));
+            DefineState(
+                AIStateIds.Idle,
+                new IdleState(this),
+                AIStateIds.Idle,
+                AIStateIds.Idle);
         }
 
         public override void Persist(StatePersister reader)
--- a/src/OpenSage.Game/Logic/AI/AIStates/AttackAreaState.cs
+++ b/src/OpenSage.Game/Logic/AI/AIStates/AttackAreaState.cs
@@ -33,8 +33,9 @@ internal sealed class AttackAreaStateMachine : StateMachineBase
 {
     public AttackAreaStateMachine(AIUpdateStateMachine parentStateMachine) : base(parentStateMachine)
     {
-        AddState(IdleState.StateId, new IdleState(this));
-        AddState(10, new AttackState(this));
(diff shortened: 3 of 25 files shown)


src/OpenSage.Game/Logic/Object/Update/AIUpdate/JetAIUpdate.cs
```diff
@@ -338,10 +338,15 @@ public JetAIUpdateStateMachine(GameObject gameObject, IGameEngine gameEngine, Je
...
         AIUpdate = aiUpdate;
 
-        AddState(WaitForAirfieldWinchesterState.StateId, new WaitForAirfieldWinchesterState(this));
+        DefineState(JetAIStateIds.CirclingDeadAirfield, new WaitForAirfieldWinchesterState(this), AIStateIds.Idle, AIStateIds.Idle);
     }
 }
 
+internal static class JetAIStateIds
+{
+    public static readonly StateId CirclingDeadAirfield = new(1013);
```
charliefoxtwo, line 347: I thought this state was specifically for when a jet has no ammunition and no airfield, and there's a different state for when it does have ammunition - hence why I named this one [Winchester](https://en.m.wikipedia.org/wiki/Multiservice_tactical_brevity_code).

charliefoxtwo (approved):

src/OpenSage.Game/Logic/Object/Update/AIUpdate/JetAIUpdate.cs
tgjones: I did wonder about the "Winchester" name, so thanks for that link - TIL :)

I just copied the name from here:

https://github.com/electronicarts/CnC_Generals_Zero_Hour/blob/0a05454d8574207440a5fb15241b98ad0b435590/GeneralsMD/Code/GameEngine/Source/GameLogic/Object/Update/AIUpdate/JetAIUpdate.cpp#L84

> and there's a different state for when it does have ammunition

Does one of the other states look like the one you had in mind? Did you mean when it does have ammunition but no airfield?

charliefoxtwo: Welp, I just realized I never actually committed all of the JetAIUpdate states I had parsed out -_-

I locally named 1012 `WaitForAirfieldState`, as this is the state where the aircraft _has_ ammunition but does not have an airfield to return to (which it will attempt to do automatically after the unit's idle timer has expired).

I was actually in the process of implementing a lot of these states a while ago, but then I got distracted by making larger changes to the state machine to accommodate... and they just never got committed.

Jets have the following states (that I was able to find):
| id |  description |
| - | - |
| `1001` | Unit has spawned in, but still needs to travel forward to its parking position |
| `1002` | Unit has been ordered to take off, but other aircraft are currently in `1003` and have reserved all active runways  |
| `1003` | Unit is taxiing to a runway for takeoff |
| `1004` | Unit has afterburners lit and is preparing for takeoff. This is sort of an "engine runup" phase, but it's complex - aircraft can also be in this phase while another aircraft in front of it is also in this phase - in this case, they'll run up short of the takeoff point, sometimes while still turning to align with the runway |
| `1005` | Unit is taking off or has just taken off and is climbing |
| `1006` | Unit is waiting to land, but other units are already in `1007` and all runways are reserved |
| `1007` | Unit is approaching the airfield to land (final approach, descending) |
| `1008` | Unit has landed, come to a full stop, and is now taxiing to parking |
| `1009` | Unit has taxied to its parking spot, and now needs to rotate 180 degrees for departure |
| `1010` | ??? |
| `1011` | Unit was ordered to land at an airfield and is flying towards the airfield |
| `1012` | Unit is idle and searching for an airfield (and has ammunition) |
| `1013` | Unit is idle and searching for an airfield (but has no ammunition) |

tgjones: Very nice! It [looks like](https://github.com/electronicarts/CnC_Generals_Zero_Hour/blob/0a05454d8574207440a5fb15241b98ad0b435590/GeneralsMD/Code/GameEngine/Source/GameLogic/Object/Update/AIUpdate/JetAIUpdate.cpp#L81C2-L81C13) `1010` is `RELOAD_AMMO` - which probably happens quickly enough that you never see it in `.sav` files.

